PR #15, opened 2015-10-23
[VGNWB-68] Register and Singup button CSS fix on Header

shivamprakash:
@erickpeirson This is how button are looking on Header. Let me know if its fine or require some changes?
screen shot 2015-10-23 at 9 40 51 am


erickpeirson:
@shivamprakash Looks great; good work!

PR #17, opened 2016-01-29
[VGNWB-86] updated header.html

jasdeepbhalla:
VGNWB-86 Task completed


erickpeirson:
Looks great! Ideally we would use reverse resolution of urls rather than hard coding the relative urls as we have done. We should create a story to go through and change all of these urls to use reverse resolution.

PR #18, opened 2016-02-01
Sprint6 vgnwb84

ndondeti:
There are 3 separate commits for this story.


annotations/views.py
```diff
@@ -158,6 +158,14 @@ def user_settings(request):
     })
     return HttpResponse(template.render(context))
 
+@login_required
```
erickpeirson: The view should be publicly available. Right now, only logged-in users can see it.

erickpeirson:
@ndondeti @allaparthi92 See my comment, above. I made a couple of commits on your branch to get the review app working; be sure to pull before making changes. Since the fix needed should be a minor change, I'll leave the PR open -- go ahead and make the modification and push to your branch, and it will update the PR.

Good work!

ndondeti:
I have made the code changes from the review.

erickpeirson:
@ndondeti and @allaparthi92 looks great! If you didn't look already, there is an auto-deployed review app at https://vogonweb-pr-18.herokuapp.com/, where you can see your changes. I'll leave this PR open so that I can demonstrate it in class.

PR #19, opened 2016-02-01
VGNWB-85

kardanitin:
We added a new html for text summary for non-logged-in users.
The existing view for text is modified.


erickpeirson:
@kardanitin @apchandak24 Your PR is backwards! If you're totally done with the story, you want to merge your branch (`sprint1-VGNWB-85`) into `develop`. Create a new PR, select `develop` as base, and select your branch for "compare".

PR #20, opened 2016-02-02
Sprint1 vgnwb 80

ndondeti:
The current state of the story is
- Added Contributors page 
- Listing the existing users but not all of their details

### DIFF
--- a/annotations/views.py
+++ b/annotations/views.py
@@ -218,6 +218,33 @@ def list_texts(request):
     }
     return HttpResponse(template.render(context))
 
+def list_users(request):
+    """
+    List all the users of Vogon web
+    """
+    queryset = VogonUser.objects.all()
+    template = loader.get_template('annotations/contributors.html')
+    userList = []
+    # text_list = Text.objects.all()
+    for user in queryset:
+        userList.append(user)
+    paginator = Paginator(userList, 25)
+
+    page = request.GET.get('page')
+    try:
+        users = paginator.page(page)
+    except PageNotAnInteger:
+        # If page is not an integer, deliver first page.
+        users = paginator.page(1)
+    except EmptyPage:
+        # If page is out of range (e.g. 9999), deliver last page of results.
+        users = paginator.page(paginator.num_pages)
+
+    context = {
+        'user_list': users,
+        'user': request.user,
+    }
+    return HttpResponse(template.render(context))
 
 @login_required
 def collection_texts(request, collectionid):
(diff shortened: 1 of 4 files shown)


annotations/views.py
```diff
@@ -218,6 +218,33 @@ def list_texts(request):
...
 
+def list_users(request):
+    """
+    List all the users of Vogon web
+    """
+    queryset = VogonUser.objects.all()
+    template = loader.get_template('annotations/contributors.html')
+    userList = []
+    # text_list = Text.objects.all()
+    for user in queryset:
```
erickpeirson: @ndondeti @allaparthi92 This is a great start! Here on line 229, however, is one potential problem.

In Django, we construct and manipulate [QuerySets](https://docs.djangoproject.com/en/1.9/ref/models/querysets/) in advance of any database calls, and then when we're good and ready (e.g. when we want to render the template, or do something fun with the results) we evaluate the QuerySet and hit the database. This is great, because it lets us set up sequential filtering and such without asking the database for fifty bajillion rows; Django sets up the query logic and then (almost always) farms out the heavy lifting to the database backend (Postgres, in this case). 

Pagination is part of the process of whittling down the QuerySet to a manageable size, not just for the user but for our application as well! If we have 10,000 users, we don't want to load all 10,000 rows from the user table and then just display 25 of them. It's better to get the 25 that we want. So we use Paginator to refine our QuerySet before it gets evaluated.

**BUT** if you iterative over a QuerySet, it will be evaluated. That means that, right here on line 229, you're making a database call with the QuerySet created on line 225: _all_ of the VogonUsers. I.e. `SELECT * FROM TABLE annotations_vogonuser;`. Yikes! And then you're passing to the Paginator not a QuerySet, but a list of VogonUser instances. The Paginator won't complain, but your users will when they have to wait five minutes for the page to load.

Take a look at the excellent [QuerySet documentation](https://docs.djangoproject.com/en/1.9/ref/models/querysets/) for more information.

erickpeirson:
Hey guys, take a look at my comment, above. Should be a quick fix; let me know if you have any questions. Since this is a PR into your own branch (just for code-review purposes), I'll let you guys go ahead and merge it yourselves.

erickpeirson:
(by the way, your code looks pretty clean! Make sure that you don't exceed 79 characters per line, even with comments)

erickpeirson:
@ndondeti  and @allaparthi92 don't forget to merge this PR.

PR #22, opened 2016-02-03
Task VGNWB-88

erhkamal:
Hi Erick,

Kindly review my changes for VGNWB-88 and give your comments.

Thanks and best regards,
Kamal

### DIFF
--- a/annotations/admin.py
+++ b/annotations/admin.py
@@ -36,10 +36,21 @@ class VogonUserAdmin(UserAdmin):
 class TextAdmin(admin.ModelAdmin):
     list_display = ('uri', 'title', 'created')
 
+class RelationAdmin(admin.ModelAdmin):
+    list_display = ('ID', 'OCCURS_IN', 'CREATED_BY', 'CREATED_AT')
+    def ID(self, obj):
+        return obj.id
+    def OCCURS_IN(self, obj):
+        return obj.occursIn.title
+    def CREATED_BY(self, obj):
+        return obj.createdBy
+    def CREATED_AT(self, obj):
+        return obj.created
+
 
 admin.site.register(VogonUser, VogonUserAdmin)
 admin.site.register(Appellation)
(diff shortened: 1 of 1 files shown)


annotations/admin.py
```diff
@@ -36,10 +36,21 @@ class VogonUserAdmin(UserAdmin):
 class TextAdmin(admin.ModelAdmin):
     list_display = ('uri', 'title', 'created')
 
+class RelationAdmin(admin.ModelAdmin):
+    list_display = ('ID', 'OCCURS_IN', 'CREATED_BY', 'CREATED_AT')
+    def ID(self, obj):
```
erickpeirson, line 41: @erhkamal319 Great start! I wonder, though, why you define all of these methods on the RelationAdmin? For most of these, you can simply use the fields on the model itself (almost always) -- e.g. `'id', 'created'`. For ForeignKey fields (e.g. `occursIn`, `createdBy`), try adding those directly, too: if I'm not mistaken, Django will evaluate the `__unicode__()` method on the target model, and return that as the value to display. Give it a try!

Oh, and try to avoid all-caps attributes/methods on classes. They only place we really use all-caps on class attributes is for constants like choices for a CharField.

annotations/admin.py
```diff
@@ -36,10 +36,21 @@ class VogonUserAdmin(UserAdmin):
 class TextAdmin(admin.ModelAdmin):
     list_display = ('uri', 'title', 'created')
 
+class RelationAdmin(admin.ModelAdmin):
```
erickpeirson, line 39: Oh, and a styling thing: always leave two blank lines above a top-level class or function declaration. See [this section in PEP8](https://www.python.org/dev/peps/pep-0008/#blank-lines).

erickpeirson:
@erhkamal319 See my comments, above. Since this is a PR into your own story branch (just for code review), I'll let you merge it yourself. Let me know if you have any questions. :-)

PR #24, opened 2016-02-03
Sprint1 vgnwb 85

kardanitin:
Added separate HTML for text view incase the user is not logged-in.
Modified the existing view to display text detail to include this use case.

### DIFF
--- a/annotations/views.py
+++ b/annotations/views.py
@@ -190,7 +190,6 @@ def network(request):
     return HttpResponse(template.render(context))
 
 
-@login_required
 def list_texts(request):
     """
     List all of the texts that the user can see, with links to annotate them.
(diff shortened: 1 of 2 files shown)


annotations/views.py
```diff
@@ -251,28 +250,35 @@ def collection_texts(request, collectionid):
 
 
 @ensure_csrf_cookie
-@login_required
 def text(request, textid):
     """
-    Provides the main text annotation view.
+    Provides the main text annotation view for logged-in users.
+    Provides summary of the text for non-logged-in users.
```
erickpeirson, line 256: Great! Good thinking to update the docstring.

annotations/views.py
```diff
@@ -251,28 +250,35 @@ def collection_texts(request, collectionid):
...
     """
-    template = loader.get_template('annotations/text.html')
+    text_dict = dict()
```
erickpeirson: Can you give this variable a more informative name? Since you use `context` for the `RequestContext` on 276, maybe you could call this `context_data`? Also, the convention for instantiating dicts is `{}`. For example:

``` python
context_data = {}
```

You can populate it with all of the values on 260 - 263 in one pass, e.g.

``` python
context_data = {
    'key': value,
    ...
}
```

erickpeirson:
@kardanitin and @apchandak24 looks like a great start! We'll want to continue to refine this view, but this give us an excellent starting place. I'll leave this PR open: if you can correct the small things noted above and push your commits to this branch, I'll go ahead and merge it.

kardanitin:
@erickpeirson : I have incorporated the suggestions.

erickpeirson:
@kardanitin Thanks! Looks good. One thing to keep in mind: we usually want to keep lines to no more than 79 characters. But views.py has quite a few lines already that are too long, so we should go through and fix this systematically.

PR #28, opened 2016-02-05
created public user profile page with required information

apchandak24:
Please check the view and let us know what needs to be added.

### DIFF
--- a/annotations/templates/annotations/user_details_public.html
+++ b/annotations/templates/annotations/user_details_public.html
@@ -17,8 +17,16 @@
 				
 				<li class="list-group-item"><strong>Affiliation: </strong>{{ detail_user.affiliation }}</li>
 				<li class="list-group-item"><strong>Location: </strong>{{ detail_user.location }}</li>
-				<li class="list-group-item"><strong>Concept power URI: </strong><a href="{{ detail_user.conceptpower_uri }}">{{ detail_user.conceptpower_uri }}</a></li>
+				{% if  detail_user.conceptpower_uri  %}
+					<li class="list-group-item"><strong>Concept power URI: </strong><a href="{{ detail_user.conceptpower_uri }}">{{ detail_user.conceptpower_uri }}</a></li>
+				{%  else %}
+					<li class="list-group-item"><strong>Concept power URI: </strong>{{ detail_user.conceptpower_uri }}</li>
+				{% endif %}
+				{% if  detail_user.link  %}
 				<li class="list-group-item"><strong>Link: </strong><a href="{{ detail_user.link}}">{{ detail_user.link}}</a></li>
+				{%  else %}
+					<li class="list-group-item"><strong>Link: </strong>{{ detail_user.link }}</li>
+				{% endif %}
 				<li class="list-group-item">User has added <strong>{{textCount}}</strong> texts</li>
 				<li class="list-group-item">Contributed to <strong>{{ annotationCount }}</strong> annotations out of <strong>{{textAnnotated}}</strong> texts</li>
 			</ul>
--- a/annotations/templates/annotations/user_details_public.html
+++ b/annotations/templates/annotations/user_details_public.html
@@ -28,7 +28,7 @@
 					<li class="list-group-item"><strong>Link: </strong>{{ detail_user.link }}</li>
 				{% endif %}
 				<li class="list-group-item">User has added <strong>{{textCount}}</strong> texts</li>
-				<li class="list-group-item">Contributed to <strong>{{ annotationCount }}</strong> annotations out of <strong>{{textAnnotated}}</strong> texts</li>
+				<li class="list-group-item">Contributed to <strong>{{ relationCount }}</strong> relations out of <strong>{{textAnnotated}}</strong> texts</li>
 			</ul>
 
 		</div>
--- a/annotations/templates/annotations/user_details_public.html
+++ b/annotations/templates/annotations/user_details_public.html
@@ -1,0 +1,30 @@
+{% extends "annotations/base.html" %}
+{% load staticfiles %}
+
+
+{% block main %}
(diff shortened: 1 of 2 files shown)


annotations/templates/annotations/user_details_public.html
```diff
@@ -0,0 +1,30 @@
...
+      	<div class="panel-body">
+      		<img src="" width="200px" height="200px"></img>
+      		<br>
+      		<br>
+			<ul class="list-group">
+				
+				<li class="list-group-item"><strong>Affiliation: </strong>{{ detail_user.affiliation }}</li>
+				<li class="list-group-item"><strong>Location: </strong>{{ detail_user.location }}</li>
+				<li class="list-group-item"><strong>Concept power URI: </strong>{{ detail_user.conceptpower_uri }}</li>
+				<li class="list-group-item"><strong>Link: </strong>{{ detail_user.link}}</li>
```
erickpeirson: Let's make this a real link (`<a href="...`)

annotations/views.py
```diff
@@ -685,16 +685,43 @@ def add_text_to_collection(request, *args, **kwargs):
...
+        })
+    else:
+        textCount = Text.objects.filter(addedBy=user).count()
+        annotationsSet = Text.objects.filter(annotators=user).distinct()
+        annotationCount = 0
+        for text in annotationsSet:
+            annotationCount += text.annotationCount
+        textAnnotated = len(annotationsSet)
+        template = loader.get_template('annotations/user_details_public.html')
+	print '------>',annotationCount
```
erickpeirson: Don't forget to remove `print` statements

annotations/templates/annotations/user_details_public.html
apchandak24: ok

annotations/views.py
```diff
@@ -685,16 +685,43 @@ def add_text_to_collection(request, *args, **kwargs):
     return JsonResponse({})
 
 
-@login_required
 def user_details(request, userid, *args, **kwargs):
+    """
+
+    :param request: The user request to fetch user details
```
erickpeirson: For arguments in docstrings, please use the format:

```
Parameters
----------
param : type
    Detailed description.

Returns
-------
type
    Detailed description.
```

annotations/views.py
```diff
@@ -685,16 +685,43 @@ def add_text_to_collection(request, *args, **kwargs):
...
+    :return: HttpResponse
+    """
     user = get_object_or_404(VogonUser, pk=userid)
-    template = loader.get_template('annotations/user_details.html')
-    context = RequestContext(request, {
-        'user': request.user,
-        'detail_user': user,
-    })
-    return HttpResponse(template.render(context))
+    if request.user.is_authenticated():
```
erickpeirson: Why does it matter whether or not the user is authenticated?

apchandak24: We have modified existing view which required login to display user details i.e his own profile.
In case the user is not loggedIn the public profile page is displayed which is a new HTML page

annotations/views.py
```diff
@@ -685,16 +685,43 @@ def add_text_to_collection(request, *args, **kwargs):
...
+            'user': request.user,
+            'detail_user': user,
+        })
+    else:
+        textCount = Text.objects.filter(addedBy=user).count()
+        annotationsSet = Text.objects.filter(annotators=user).distinct()
+        annotationCount = 0
+        for text in annotationsSet:
+            annotationCount += text.annotationCount
+        textAnnotated = len(annotationsSet)
```
erickpeirson: Yikes! This will cause a pretty big database query. When you take the `len()` of a `QuerySet`, it will evaluate the QuerySet (yielding a list of objects) first. That means that you are executing the query `SELECT * FROM TABLE annotations_text;`, and then evaluating the length of the resulting list of objects. If you just want to get the number of records, use the QuerySet's `count()` method (e.g. `annotationSet.count()`).  This will run `SELECT COUNT(*)...` instead.

Take a look at the [QuerySet documentation](https://docs.djangoproject.com/en/1.9/ref/models/querysets/).

apchandak24: yes

annotations/views.py
erickpeirson: Ah, gotcha. That's clever

annotations/views.py
```diff
@@ -685,14 +685,46 @@ def add_text_to_collection(request, *args, **kwargs):
     return JsonResponse({})
 
 
-@login_required
 def user_details(request, userid, *args, **kwargs):
+    """
+    Provides users with their own profile view and public profile view of other users in case they are loggedIn.
```
erickpeirson, line 690: Watch your line length: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

annotations/views.py
```diff
@@ -685,14 +685,46 @@ def add_text_to_collection(request, *args, **kwargs):
...
+        context = RequestContext(request, {
+            'user': request.user,
+            'detail_user': user,
+        })
+    else:
+        textCount = Text.objects.filter(addedBy=user).count()
+        annotationsSet = Text.objects.filter(annotators=user).distinct()
+        annotationCount = 0
+        for text in annotationsSet:
+            annotationCount += text.annotationCount
```
erickpeirson: We can simplify this quite a bit by using the `RelatedManager` for `Annotations` on the `VogonUser` class. Take a look at the `Annotation` model: since there is a `ForeignKey` to `VogonUser` with no explicit `related_name`, there should automatically be a `RelatedManager` on `VogonUser` called `annotation_set`. So instead of iterating over all of the texts that the user has annotated, you can just do:

``` python
annotationCount = user.annotation_set.count()
```

That **should** work, anyway. Give it a try!

erickpeirson:
@apchandak24 Looks great. Good work!

PR #29, opened 2016-02-05
VGNWB-96 Replaced list with table, text sorted by title

erhkamal:
Hi Erick,

Could you please review my changes and comment?

Thanks and best regards,
Kamal

### DIFF
--- a/annotations/static/annotations/css/main.css
+++ b/annotations/static/annotations/css/main.css
@@ -268,3 +268,11 @@ circle.node {
 .navbar-nav {
     font-size: 1.2em;
 }
+
+.text-table:hover{
+    background-color: #F0F0F0;
+}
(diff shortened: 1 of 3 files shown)


annotations/static/annotations/css/main.css
```diff
@@ -268,3 +268,11 @@ circle.node {
 .navbar-nav {
     font-size: 1.2em;
 }
+
+.text-table:hover{
```
erickpeirson, line 272: Why not use [.table-hover](http://getbootstrap.com/css/#tables-hover-rows) ?

erickpeirson:
@erhkamal319 Looks good, except that I don't think that you need the extra CSS classes to achieve the look that you want

PR #30, opened 2016-02-08
Sprint2 vgnwb 80

ndondeti:
Incorporates paging and sorting on the list of users.

### DIFF
--- a/annotations/views.py
+++ b/annotations/views.py
@@ -218,7 +218,7 @@ def list_texts(request):
     }
     return HttpResponse(template.render(context))
 
-@login_required
+
 def list_user(request):
     """
     List all the users of Vogon web
--- a/annotations/views.py
+++ b/annotations/views.py
@@ -226,14 +226,11 @@ def list_user(request):
 
     template = loader.get_template('annotations/contributors.html')
 
-    #Default Sort
-    sort = "user_name"
     #To map the column sort parameter from UI to the column name in DB
     sort_dict = {"user_name":"username", "name":"full_name",
      "aff":"affiliation", "loc":"location"}
 
-    if(request.GET.get('sort') != None):
-        sort = request.GET.get('sort')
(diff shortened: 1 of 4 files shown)


annotations/views.py
```diff
@@ -218,6 +218,44 @@ def list_texts(request):
...
+    """
+    List all the users of Vogon web
+    """
+
+    template = loader.get_template('annotations/contributors.html')
+
+    #Default Sort
+    sort = "user_name"
+    #To map the column sort parameter from UI to the column name in DB
+    sort_dict = {"user_name":"username", "name":"full_name",
```
erickpeirson, line 230: Why not just use the database field names to begin with?

annotations/views.py
```diff
@@ -218,6 +218,44 @@ def list_texts(request):
...
+     "aff":"affiliation", "loc":"location"}
+
+    if(request.GET.get('sort') != None):
```
erickpeirson: A simpler way to do this would be to provide the alternate (default) value to `GET.get()`, e.g.:

``` python
sort_column = request.GET.get('sort', 'user_name')    # If no 'sort' key, then returns ``'user_name'``
```

annotations/views.py
ndondeti, line 230: I would but that would be exposing the the database structure to the outside world. In this case it is not going to affect but in general its considered bad practice to do so. The mapping is just to be on the safer side of things.

annotations/views.py
ndondeti: Will incorporate this change. Do you need us to merge this with the story branch?

annotations/views.py
```diff
@@ -218,6 +218,44 @@ def list_texts(request):
     }
     return HttpResponse(template.render(context))
 
+@login_required
```
erickpeirson: This view should be available to the public

PR #35, opened 2016-02-10
[VGNWB-82]: User profile image

nravich2:
Code change and testing completed under development environment.Please review the changes.
Thanks


vogon/local_settings.py
```diff
@@ -129,7 +129,10 @@
 # https://docs.djangoproject.com/en/1.8/howto/static-files/
 
 STATIC_URL = '/static/'
-# STATIC_ROOT = '/diging/wwwnginx/vogon/static/'
+STATIC_ROOT = '/diging/wwwnginx/vogon/static/'
+MEDIA_ROOT = '/home/jasdeep/Documents'
```
erickpeirson: Don't hard-code this; you should use the `os` package to get the current working directory, and then construct the path relative to that.

e.g. 

``` python
PROJECT_ROOT = os.path.dirname(os.path.abspath(__file__))
MEDIA_ROOT = os.path.join(PROJECT_ROOT, 'media')
```

Otherwise....
image

jasdeepbhalla:
This is the new and updated pull request!

annotations/views.py
```diff
@@ -139,13 +139,19 @@ def user_settings(request):
     """ User profile settings"""
 
     if request.method == 'POST':
-        form = UserChangeForm(request.POST)
+        form = UserChangeForm(request.POST,request.FILES)
```
erickpeirson: Always leave a space after a comma

annotations/views.py
```diff
@@ -139,13 +139,19 @@ def user_settings(request):
...
         if form.is_valid():
-            for field in ['first_name', 'last_name', 'email']:
+            
+            
+            for field in ['first_name', 'last_name', 'email','imagefile']:
```
erickpeirson: Space after comma

annotations/views.py
```diff
@@ -139,13 +139,19 @@ def user_settings(request):
...
                 value = request.POST.get(field, None)
                 if value:
                     setattr(request.user, field, value)
             request.user.save()
+
+                #to save uploaded file
```
erickpeirson: Block comments should always be on the same level of indentation as the code that they are about. The hash should be followed by a space, and the comment should be grammatically correct: e.g. start with a caps letter, end with a period.

annotations/views.py
```diff
@@ -158,6 +164,13 @@ def user_settings(request):
     })
     return HttpResponse(template.render(context))
 
+def save_file(f):
```
erickpeirson: Needs a docstring

annotations/views.py
```diff
@@ -158,6 +164,13 @@ def user_settings(request):
...
+    filename = f.name
+    fd = open('%s/%s' % (settings.MEDIA_ROOT, str(filename)), 'wb')
+    for chunk in f.chunks():
+        fd.write(chunk)
+    fd.close()
+
 def about(request):
```
erickpeirson, line 199: Add another newline, so that the function below has two empty lines above it.

annotations/models.py
```diff
@@ -47,6 +47,7 @@ class VogonUser(AbstractBaseUser, PermissionsMixin):
     link = models.URLField(max_length=500, blank=True, null=True)
     full_name = models.CharField(max_length=255, blank=True, null=True)
     conceptpower_uri = models.URLField(max_length=500, blank=True, null=True)
+    imagefile = models.FileField(upload_to='documents/%Y/%m/%d',blank=True, null=True,default='')
```
erickpeirson: Spaces after commas. Don't go past 79 characters; break the line like this:

``` python
    imagefile = models.FileField(upload_to='documents/%Y/%m/%d',
                                 blank=True, null=True, default='')
```

erickpeirson:
Great work! See my line comments, above. Also, you'll need to resolve conflicts before this can be merged.

annotations/views.py
```diff
@@ -183,6 +190,25 @@ def user_settings(request):
     })
     return HttpResponse(template.render(context))
 
+
+def save_file(f):
```
erickpeirson: Hmm, did you look at the [File Uploads](https://docs.djangoproject.com/en/1.9/topics/http/file-uploads/#handling-uploaded-files-with-a-model) documentation? The file should be saved when the form is saved, so you may not need this function at all.

annotations/views.py
```diff
@@ -164,13 +165,19 @@ def user_settings(request):
...
+        form = UserChangeForm(request.POST, request.FILES)
         if form.is_valid():
-            for field in ['first_name', 'last_name', 'email']:
+            
+            
+            for field in ['first_name', 'last_name', 'email', 'imagefile']:
                 value = request.POST.get(field, None)
                 if value:
                     setattr(request.user, field, value)
             request.user.save()
```
erickpeirson: Why are we saving the user directly, and not just saving the form?

nravich2:
Worked on the comments. Upload functionality done using form.save() method.

erickpeirson:
Excellent! Can I ask, why did you add 9 new migrations?

nravich2:
Initially , when we tried form.save() , we thought there was a problem with the 'upload_to' argument in the VogonUser Model. So we changed the model with various values for that argument. 
Only later , we found that the images were actually getting saved and the problem is caused due to some other issue.

erickpeirson:
@nravich2 Ok, but there is no need for all of those migrations. You should delete all of the migrations that you created, and then run makemigrations again to create a single migration.

nravich2:
@erickpeirson : Single migration created as required.

erickpeirson:
@nravich2 Awesome, thanks! I'm getting some strange behavior, though; after I upload a file, and then go back to my settings, I see a link that looks like this:

image

That points to: http://127.0.0.1:8000/Users/erickpeirson/Projects/vogon-web/vogon/media/IMG_2887.jpg , which raises a 404. 

I'm also noticing that you have a hard-coded full path in the migration. Taken together, these lead me to believe that you are passing a full path rather than a relative path to the `upload_to` parameter in the `FileField` constructor on the model. The value of `upload_to` should be a relative path, e.g. `"userimages/"`, which is relative to the `MEDIA_ROOT`.

nravich2:
Changed to relative path in settings file. Added url to serve media files in localhost

erickpeirson:
@nravich2 Ok, sorry for the confusion about the URLs. It's all coming back to me now!

In production, it is not recommended (and, in our case, not possible) to serve user-uploaded files via Django. Django is not very secure when it comes to serving static content, and in our case we don't have a persistent filesystem in production: the Heroku workers come and go. So the filesystem approach that you're working with to develop this feature works for development, but we'll need to alter it slightly to accommodate an external solution for media storage.

For testing purposes, however, we can serve the media files via Django, and this does require using a URL pattern (as you discovered). I'd suggest doing it like this:

``` python
if settings.DEBUG:
    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

There are two reasons for this: 
1. It will prevent insecure behavior in production, and
2. This takes into account the `MEDIA_URL` setting, which would be ignored if we hard-coded the URL pattern.

Make sense? This afternoon I will get clear about the production file handling strategy, and we can wrap this up. Sorry it has been such a long road, but you guys have done a great job!

nravich2:
@erickpeirson - Thanks ! Yes , it makes sense. Also modifying urls.py this way (using MEDIA_URL) , the code is much cleaner. I have done the changes.Please have a look and let me know.

erickpeirson:
Going to keep this PR open until we get the S3 bit implemented. But this looks great, good work!

nravich2:
Uploading of profile image in S3 complete. The URL of the image is saved into the Vogon User model.Please check and let us know.

nravich2:
Uploading of profile image in S3 complete. The URL of the image is saved into the Vogon User model.Please check and let us know.

annotations/templates/annotations/settings.html
```diff
@@ -1,32 +1,71 @@
...
+        alert("No file selected.");
+        return;
+    }
+    get_signed_request(file);
+}
+/*
+   Bind listeners when the page loads.
+*/
+(function() {
+    document.getElementById("file_input").onchange = init_upload;
```
erickpeirson: image

Are you sure that this is being executed after the page loads? You might consider using jQuery for this:

``` javascript
$('body').ready(function() {
    $('#file_input').on('change', init_upload);
});
```

annotations/templates/annotations/settings.html
```diff
@@ -1,32 +1,71 @@
...
+      <label for="location">Location</label>
+      <input type="text" class="form-control" id="id_location" value="{{ location }}">
+      </div>
+      <div class="form-group">
+      <label for="link">Link</label>
+      <input type="text" class="form-control" id="id_link" value="{{ link }}">
+      </div>
+      <div class="form-group">
+      <label for="imagefile">Please upload an image</label>
+      <input type="file" id="file_input"/>
```
erickpeirson: If this is inside the `<form></form>`, then the file will be uploaded to VogonWeb, too. We don't want that. See the first code-cell in this section: https://devcenter.heroku.com/articles/s3-upload-python#setting-up-the-client-side-code

annotations/models.py
```diff
@@ -50,6 +51,8 @@ class VogonUser(AbstractBaseUser, PermissionsMixin):
     link = models.URLField(max_length=500, blank=True, null=True)
     full_name = models.CharField(max_length=255, blank=True, null=True)
     conceptpower_uri = models.URLField(max_length=500, blank=True, null=True)
+    imagefile = models.FileField(upload_to='uploads/',
```
erickpeirson: Since we're no longer storing an image file locally, this should be a `URLField`.

annotations/views.py
```diff
@@ -167,21 +168,32 @@ def user_settings(request):
     """ User profile settings"""
 
     if request.method == 'POST':
-        form = UserChangeForm(request.POST, instance=request.user)
+        form = UserChangeForm(request.POST, request.FILES,instance=request.user)
```
erickpeirson: Now that we have S3 upload, we shouldn't be handling any files here. Just the URL for the image on S3, which should be in the `POST` data.

annotations/templates/annotations/settings_old.html
```diff
@@ -0,0 +1,33 @@
+{% extends "annotations/base.html" %}
```
erickpeirson: Do we still need this file for some reason?

erickpeirson:
@nravich2 Great work!! There are still a few things to do; see my comments, and let me know if you have any questions.

nravich2:
All our changes are in VGNWB130 and we want to have a PR into VGNWB82 from VGNWB130. I am not sure why is it pointing as develop from VGNWB82. Please let us know if we have to raise a new PR . Thanks

PR #37, opened 2016-02-10
VGNWB-106 Excluding anonmyous user

ndondeti:


annotations/views.py
```diff
@@ -240,7 +240,7 @@ def list_user(request):
 
     sort_column = sort_dict[sort]
 
-    queryset = VogonUser.objects.all().order_by(sort_column)
+    queryset = VogonUser.objects.all().exclude(id = -1).order_by(sort_column)
```
erickpeirson: No need for the `.all()` bit; [`exclude()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#exclude) is enough

PR #39, opened 2016-02-12
VGNWB-105 Changing the contributors page from list to table

ndondeti:
This pull request is for both VGNWB-105 and VGNWB-108.

### DIFF
--- a/annotations/templates/annotations/contributors.html
+++ b/annotations/templates/annotations/contributors.html
@@ -31,7 +31,7 @@
                 {% for users in user_list %}
                 <tr>
                     <td class="col-sm-3">
-                    <a href="/user/{{users.id}}/">{{ users.username }} &nbsp;{% if users.is_admin %}<span class="glyphicon glyphicon-user"></span>{% endif %}</a></td>
+                    <a href="{% url "user_details" users.id %}">{{ users.username }} &nbsp;{% if users.is_admin %}<span class="glyphicon glyphicon-user"></span>{% endif %}</a></td>
                     <td class="col-sm-3">{% if users.full_name %}{{users.full_name}}{% endif %}</td>
                     <td class="col-sm-3">{% if users.affiliation %}{{users.affiliation}}{% endif %}</td>
                     <td class="col-sm-3">{% if users.location %}{{users.location}}{% endif %}</td>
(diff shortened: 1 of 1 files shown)


annotations/templates/annotations/contributors.html
```diff
@@ -16,28 +16,34 @@
...
-
+                <tr>
                   {% if users.is_admin %}
-                    <a href="/user/{{users.id}}/" class="list-group-item col-sm-3">
-                    {{ users.username }} &nbsp;<span class="glyphicon glyphicon-user"></span></a>
+                    <td href="/user/{{users.id}}/" class="col-sm-3">
+                    <a href="/user/{{users.id}}/">{{ users.username }} &nbsp;<span class="glyphicon glyphicon-user"></span></a></td>
                   {% else %}
-                  <a href="/user/{{users.id}}/" class="list-group-item col-sm-3">{{ users.username }}</a>
+                  <td  class="col-sm-3"><a href="/user/{{users.id}}/">{{ users.username }}</a></td>
```
erickpeirson: We can tighten this up a bit. 

First, use reverse URL resolution rather than hard-coding the URLs. See [this page](https://docs.djangoproject.com/en/1.9/topics/http/urls/#reverse-resolution-of-urls). For example:

`{% url "user_details" user.id %}`

Second, Don't Repeat Yourself. You declare the `td` element twice; the `if` statement is only about wheher or not to show the glyphicon, so the glyphicon bit should be the only part inside of the `if` statement.

erickpeirson:
@ndondeti @allaparthi92 Some minor tweaks, but otherwise looks great!

annotations/templates/annotations/contributors.html
```diff
@@ -16,28 +16,30 @@
...
-                    {{ users.username }} &nbsp;<span class="glyphicon glyphicon-user"></span></a>
-                  {% else %}
-                  <a href="/user/{{users.id}}/" class="list-group-item col-sm-3">{{ users.username }}</a>
-                  {% endif %}
-                  <div class="list-group-item col-sm-3">{{users.full_name}}</div>
-                  <div class="list-group-item col-sm-3">{{users.affiliation}}</div>
-                  <div class="list-group-item col-sm-3">{{users.location}}</div>
+                <tr>
+                    <td class="col-sm-3">
+                    <a href="/user/{{users.id}}/">{{ users.username }} &nbsp;{% if users.is_admin %}<span class="glyphicon glyphicon-user"></span>{% endif %}</a></td>
```
erickpeirson: This is still hard-coded; use the `{% url ... %}` tag

erickpeirson:
@ndondeti and @allaparthi92  great work!

PR #40, opened 2016-02-12
VGNWB:93:Add sorting functionality on click of column header in /text…

harshaneelhg:


annotations/templates/annotations/list_texts.html
```diff
@@ -67,4 +67,19 @@
...
+  sessionStorage.setItem('order', 'created');
+});
+function getQueryString(q){
+  page = document.getElementById('page_number_'+q);
+  if(sessionStorage.getItem('order') == null)
+    page.href='?page='+q+'&order_by=title';
+  else
+    page.href='?page='+q+'&order_by='+ sessionStorage.getItem('order');
+}
+</script>
```
erickpeirson, line 84: Why are we doing this? Can't we keep track of everything in the GET parameters? What happens when the session ends?

annotations/templates/annotations/list_texts.html
```diff
@@ -45,7 +45,7 @@
                     {% endif %}
                     {% for p in texts.paginator.page_range %}
                     <li{% if p == texts.number %} class="active"{% endif %}>
-                        <a href="?page={{p}}">{{p}}</a>
+                        <a id="page_number_{{p}}" onclick=getQueryString('{{p}}')>{{p}}</a>
```
erickpeirson, line 48: No javascript, please

erickpeirson:
@harshaneelhg Good start, but we should implement this without javascript. Instead, use template tags to pass the current parameters from the view to the template. Let me know if you have questions about how to do this. :-)

harshaneelhg:
Initially we did this without using javascript. But there was an issue with that. We were getting properly sorted columns  but when we tried to change the page again default sorting was displayed. So we have used javascript to persist what sorting key must be used across different pages.

annotations/templates/annotations/list_texts.html
harshaneelhg, line 84: Yeah but when we change the page using pagination control we need to know what sort key we are using. That is what we are doing using JavaScript. JS keeps track of what sort key is to be used when we change page.

erickpeirson:
@harshaneelhg Let's talk about this today during lab. We can (and should) do this without the JS. :-)

PR #43, opened 2016-02-15
VGNWB-90 Adding message to the annotate text view

ndondeti:
screenshot from 2016-02-15 09 19 03
- Added the message in the above spot.


erickpeirson:
@ndondeti Good start, but this way it's only visible at that particular stage of annotation. For example:

image

Could we put it down here somewhere? 

image

erickpeirson:
Awesome; thanks!

PR #45, opened 2016-02-15
VGNWB-91 Added the functionality which has the ability to filter the …

allaparthi92:
Added the functionality which has the ability to filter the concept changelist by type

### DIFF
--- a/concepts/admin.py
+++ b/concepts/admin.py
@@ -11,9 +11,9 @@ resolve.verbose_name = 'resolve selected concepts'
 class ConceptAdmin(admin.ModelAdmin):
     model = Concept
     search_fields = (   'label',    )
-    list_display = (    'label', 'resolved','concept_state'  )
+    list_display = (    'label', 'resolved','concept_state','typed'  )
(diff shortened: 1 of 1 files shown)


concepts/admin.py
```diff
@@ -11,9 +11,9 @@ def resolve(modeladmin, request, queryset):
 class ConceptAdmin(admin.ModelAdmin):
     model = Concept
     search_fields = (   'label',    )
-    list_display = (    'label', 'resolved','concept_state'  )
+    list_display = (    'label', 'resolved','concept_state','typed'  )
     actions = ( resolve,    )
-    list_filter=('concept_state',)
+    list_filter=('concept_state','typed')
```
erickpeirson: Minor stylistic corrections: can you add a space after each comma? And while you're at it, can you remove the extra whitespace inside the parentheses? I know it was already like that, but we might as well clean it up while we're at it.

For example:

``` python
search_fields = ('label', )
```

erickpeirson:
@allaparthi92 One small tweak, and then this looks great! :-)

PR #47, opened 2016-02-15
Story/vgnwb 93 - Enabled sort by asc and desc on Text table columns

erhkamal:
Hi Erick,

Could you please review and make a comment?

Thanks,
Kamal & Harshaneel.

### DIFF
--- a/annotations/templates/annotations/list_texts.html
+++ b/annotations/templates/annotations/list_texts.html
@@ -22,13 +22,13 @@
         <div class="col-sm-8 col-sm-offset-2">
             {% if texts|length > 0 %}
             <table class="table table-hover">
-              <th class="col-sm-8"><a href="?order_by=title" id="order_title">Text title</a></th>
-              <th class="col-sm-3"><a href="?order_by=created" id="order_publication">Publication date</a></th>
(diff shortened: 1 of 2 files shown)


annotations/templates/annotations/list_texts.html
```diff
@@ -22,13 +22,13 @@
         <div class="col-sm-8 col-sm-offset-2">
             {% if texts|length > 0 %}
             <table class="table table-hover">
-              <th class="col-sm-8">Text title</th>
-              <th class="col-sm-3">Publication date</th>
+              <th class="col-sm-8"><a href="{% if order_by == '-title' %}?order_by=title{% else %}?order_by=-title{% endif %}" id="order_title">Text title</a></th>
```
erickpeirson: What happens when I'm on page 4, and I click on this link?

erhkamal, line 25: goes back to page 1 and toggles the sort order.

erickpeirson:
Actually, ignore that last comment. It makes sense that we would not preserve the page number here.

erhkamal:
Yeah, thanks.

PR #50, opened 2016-02-17
[VGNWB-101] Changed image circles to thumbnails and fixed image alignment problem

jasdeepbhalla:


erickpeirson:
Looks like there are some conflicts that need resolving! Let me know if you need help making sense of the conflicts -- in general you should accept what is on develop, unless it is something that you explicitly meant to change.

erickpeirson:
Also, dimensions are definitely consistent, but resolution is not. For example, see the differences in resolution among these four images:

image

This could be addressed by making the images smaller on the page, and resizing them so that they are all the same shape (and thus don't get distorted).

erickpeirson:
Awesome!

PR #51, opened 2016-02-17
Initial commit for the user search

ndondeti:


annotations/views.py
```diff
@@ -268,11 +268,16 @@ def list_user(request):
     sort_dict = {"user_name":"username", "name":"full_name",
      "aff":"affiliation", "loc":"location"}
 
+    srch_term = request.GET.get('srch-term')
```
erickpeirson: No need to neglect vowels, `search_term` and `search-term` are much clearer and don't make me feel like my brain is getting squeezed in a vice ;-)

annotations/views.py
```diff
@@ -268,11 +268,16 @@ def list_user(request):
...
     sort = request.GET.get('sort', 'user_name')
 
     sort_column = sort_dict[sort]
 
-    queryset = VogonUser.objects.exclude(id = -1).order_by(sort_column)
+    if(srch_term):
+        queryset = VogonUser.objects.filter(username__contains = srch_term).exclude(id = -1).order_by(sort_column)
```
erickpeirson: Might want to use `__icontains` rather than `__contains` (see [this page](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#icontains)), since mysterious things can happen with capitalization in GET requests...

annotations/views.py
```diff
@@ -268,11 +268,16 @@ def list_user(request):
...
+    else:
+        queryset = VogonUser.objects.exclude(id = -1).order_by(sort_column)
```
erickpeirson: DRY. The only difference between this and line 277 is the `filter()`.

annotations/views.py
```diff
@@ -286,6 +291,7 @@ def list_user(request):
         users = paginator.page(paginator.num_pages)
 
     context = {
+        'srch_term' : srch_term,
```
erickpeirson: image

erickpeirson:
Great start! See line comments

annotations/views.py
ndondeti: Erick, how should I go about this. Should I create a regex for filtering? Because for contains I cannot use any default value.

ndondeti:
Hey Erick, I hope that all the users have full name by default. If they do not have a full name then they wont appear in to default contributors page. If this has to be avoided then the if condition need to remain.

erickpeirson:
@ndondeti I don't follow; why is that?

erickpeirson:
@ndondeti Since this is a sprint PR, you can merge this whenever you like. Or we can leave it open if that's helpful. :-)

PR #54, opened 2016-02-19
Vgnwb 107

ndondeti:


annotations/views.py
```diff
@@ -268,12 +268,14 @@ def list_user(request):
     sort_dict = {"user_name":"username", "name":"full_name",
```
erickpeirson, line 268: Sorry if we talked about this before: why do we need a separate set of names for fields? Can't we just use the field names on the model?

annotations/views.py
```diff
@@ -268,12 +268,14 @@ def list_user(request):
...
      "aff":"affiliation", "loc":"location"}
 
+    search_term = request.GET.get('search_term', "")
     sort = request.GET.get('sort', 'user_name')
 
     sort_column = sort_dict[sort]
 
-    queryset = VogonUser.objects.exclude(id = -1).order_by(sort_column)
-    paginator = Paginator(queryset, 25)
+    queryset = VogonUser.objects.filter(full_name__icontains = search_term).exclude(id = -1).order_by(sort_column)
```
erickpeirson: Why not set the default value on line 271 to `None`, and then only apply the filter if it is set? You can keep it DRY by doing:

``` python
queryset = VogonUser.objects.exclude(id=-1).order_by(sort_column)
if search_term:
    queryset = queryset.filter(full_name__icontains=search_term)
```

This way, if there is no search, we see all of the users (except AnonymousUser).

erickpeirson:
Looks good, works great! A couple of minor tweaks, above, but otherwise ready to merge.

annotations/views.py
ndondeti, line 268: The reason is we do not want to expose our table structure to the outside world. It might seem trivial but you never know when it might be used against you. Its just a good practice I read somewhere. I can change it if you feel otherwise. :-)

erickpeirson, line 268: Ah, that makes sense. Fine with me!

PR #55, opened 2016-02-19
VGNWB-110 bug fix

apchandak24:
Bug fix for showing home page instead of dashboard when user is logged in


erickpeirson:
Great! Simple enough

PR #56, opened 2016-02-19
Adding validations for full_name, link, location, affliation. making …

kardanitin:
…them mandatory

### DIFF
--- a/annotations/forms.py
+++ b/annotations/forms.py
@@ -122,3 +122,19 @@ class UserChangeForm(forms.ModelForm):
         model = VogonUser
         fields = ('full_name', 'email', 'affiliation', 'location', 'link', )
 
+    def clean(self):
+        validationErrorDict = {}
+        if not self.cleaned_data.get('full_name'):
+            validationErrorDict['full_name']= ValidationError(_('Missing full name.'), code='required')
+
+        if not self.cleaned_data.get('affiliation'):
+            validationErrorDict['affiliation']= ValidationError(_('Missing affiliation.'), code='required')
+
+        if not self.cleaned_data.get('location'):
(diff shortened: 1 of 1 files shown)


annotations/forms.py
```diff
@@ -122,3 +122,19 @@ class Meta:
         model = VogonUser
         fields = ('full_name', 'email', 'affiliation', 'location', 'link', )
 
+    def clean(self):
```
erickpeirson: `clean()` is a good place to put validation that involves field interdependencies, but since these are checking fields independently we should add a separate `clean_[fieldname]` method for each. See [Form and field validation](https://docs.djangoproject.com/en/1.9/ref/forms/validation/).

annotations/forms.py
```diff
@@ -122,3 +122,19 @@ class Meta:
...
+            validationErrorDict['full_name']= ValidationError(_('Missing full name.'), code='required')
+
+        if not self.cleaned_data.get('affiliation'):
+            validationErrorDict['affiliation']= ValidationError(_('Missing affiliation.'), code='required')
+
+        if not self.cleaned_data.get('location'):
+            validationErrorDict['location']= ValidationError(_('Missing location.'), code='required')
+
+        if not self.cleaned_data.get('link'):
+            validationErrorDict['link']= ValidationError(_('Missing link.'), code='required')
```
erickpeirson: Once we have these in their own separate `clean_` methods, we should just raise `ValidationError` -- no need to compile them.

erickpeirson:
@kardanitin This looks good! Just a little reorganization (see line comments), and we're all set.

erickpeirson:
@kardanitin Great, thanks!

PR #70, opened 2016-03-15
Vgnwb 121

yyk16293:


concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
         authorities.resolve(type(obj), obj)
 resolve.verbose_name = 'resolve selected concepts'
```
erickpeirson, line 13: Always leave two blank lines before a top-level class or function declaration

concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
...
 
+def merge_concepts(modeladmin, request, queryset):
+
```
erickpeirson: Be sure to add a docstring

concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
...
+    opts = modeladmin.model._meta
+    app_label = opts.app_label
+    RESOLVED_STATUS = 'Resolved'
+    resolvedCount = 0
+    resolvedConcept = None
+
+    for item in queryset:
```
erickpeirson: Rather than iterate over the entire queryset (which loads the whole result set from the database), we can cut down on our DB overhead by using queryset methods. E.g.

``` python
resolvedCount = queryset.filter(concept_state=Concept.RESOLVED).count()
```

See [this page](https://docs.djangoproject.com/en/1.9/ref/models/querysets/) for more information.

concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
         authorities.resolve(type(obj), obj)
 resolve.verbose_name = 'resolve selected concepts'
 
+def merge_concepts(modeladmin, request, queryset):
+
+    opts = modeladmin.model._meta
+    app_label = opts.app_label
+    RESOLVED_STATUS = 'Resolved'
```
erickpeirson: Don't hard-code this; the status options are defined on the model itself. E.g. you can use `Concept.RESOLVED`

concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
...
+        "opts": opts,
+        "app_label": app_label
+    }
+
+    #VGNWB-121 gets called only when there is only one resolved concept
+    if resolvedCount == 1:
+        return render_to_response('admin/merge_concepts_resolved.html', context)
+
+    return render_to_response('admin/merge_concepts.html', {'concepts': queryset})
+merge_concepts.verbose_name = "Merge Concepts"
```
erickpeirson: `verbose_name` shouldn't have caps in it; also, not really necessary since the method is already called `merge_concepts` (which will display as "merge concepts").

concepts/admin.py
```diff
@@ -8,16 +11,46 @@ def resolve(modeladmin, request, queryset):
...
+
+
+#def merge_concepts(self, request, queryset):
```
erickpeirson: Remove any extraneous code

erickpeirson:
@yyk16293 Great start! I made some line comments above; let me know if you have any questions.

Here's what I see when I select some concepts, one of which is resolved:

image

Here's the description of the story: "If one and only one of the selected concepts is "resolved", and the remainder are not, then we should ask the user to confirm that they want to merge the unresolved concepts into the resolved concept."

So, we can see the resolved concept, and the prompt text looks good. But we also need to see the unresolved concepts, as well.

erickpeirson:
This looks good; next we'll need to actually perform the merge; see story VGNWB-146

PR #71, opened 2016-03-17
Adding foreign key

yyk16293:


concepts/models.py
```diff
@@ -51,6 +51,7 @@ class Concept(HeritableObject):
         (RESOLVED,'Resolved'),
     )
     concept_state=models.CharField(max_length=10,choices=concept_state_choices, default='Pending')
+    merged_with = models.ForeignKey('Concept', related_name='merged_with_field', **optional)
```
erickpeirson: Need a semantically meaningful `related_name`

erickpeirson:
Whoops, looks like there are two migrations there; we only need one. In a case like this, where you made a change to the model before the PR was merged, just roll back your database (`python manage.py migrate annotations 0001``), delete the migrations that you created (``0002``and``0003``), and create a new migration.

PR #76, opened 2016-03-22
Sprint4 vgnwb 135

apchandak24:
Initial Changes for Recent Activity


annotations/models.py
```diff
@@ -159,7 +159,6 @@ class TextCollection(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     participants = models.ManyToManyField(VogonUser, related_name='contributes_to')
 
-
 class Text(models.Model):
```
erickpeirson, line 176: Need that blank line back :-)

annotations/views.py
```diff
@@ -330,6 +332,59 @@ def collection_texts(request, collectionid):
     }
     return HttpResponse(template.render(context))
 
+class RecentActivity(models.Model):
```
erickpeirson: Two blank lines above all top-level class and function declarations

annotations/views.py
```diff
@@ -330,6 +332,59 @@ def collection_texts(request, collectionid):
...
+    def __init__(self):
+        self.appellation = 0
+        self.text = 0
+
+def recent_activity(request):
```
erickpeirson, line 374: Two blank lines above all top-level class and function declarations

annotations/views.py
```diff
@@ -330,6 +332,59 @@ def collection_texts(request, collectionid):
...
+                        activity = userDict.get(username)
+                        activity.appellation +=1
+                appellation_list.remove(appellation)
+
+    context = {
+        'recent_activity': activity_data
+    }
+    return HttpResponse(template.render(context))
 
 @ensure_csrf_cookie
```
erickpeirson, line 395: Two blank lines above all top-level class and function declarations

annotations/views.py
```diff
@@ -330,6 +332,59 @@ def collection_texts(request, collectionid):
...
+    def __init__(self):
+        self.appellation = 0
+        self.text = 0
+
+def recent_activity(request):
+    from datetime import datetime, timedelta
+    from django.utils import timezone
+    template = loader.get_template('annotations/recent_activity.html')
+    time_threshold = datetime.now() - timedelta(days=1)
+    text_list = list(Text.objects.filter(added__gt=time_threshold).order_by('-added')[:10])
```
erickpeirson: So you only want to look at texts that have been added within the last day? That doesn't seem right.

kardanitin: @erickpeirson : We can change that to 7 days to consider activity for the complete week.

annotations/models.py
kardanitin, line 176: Done

annotations/views.py
kardanitin: Done

annotations/views.py
kardanitin, line 374: Done

annotations/views.py
kardanitin, line 395: Done

vogon/urls.py
```diff
@@ -52,8 +52,10 @@
     url(r'^accounts/profile/', views.dashboard, name='dashboard'),
     url(r'^accounts/settings/$', views.user_settings),
     url(r'^accounts/register/$', views.register),
+    url(r'^activity/$', views.recent_activity),
     url(r'^accounts/logout/$', 'django.contrib.auth.views.logout',
                           {'next_page': '/accounts/login/'}),
+    url(r'^activity/$', views.recent_activity),
```
erickpeirson, line 57: Duplicate

kardanitin, line 57: @erickpeirson : We have added that for testing purposes. we would have the recent activity on our homepage

annotations/views.py
```diff
@@ -364,6 +366,62 @@ def collection_texts(request, collectionid):
...
+
+    activity_data={}
+
+    if text_list:
+        for initialHour in range(1, 25):
+            for item in text_list:
+                username = item.addedBy.username
+                if not activity_data.get(initialHour):
```
erickpeirson: This works, but is semantically a little strange. Why not 

``` python
if not initialHour in activity_data:
```

?

annotations/views.py
```diff
@@ -364,6 +366,62 @@ def collection_texts(request, collectionid):
     return HttpResponse(template.render(context))
 
 
+class RecentActivity(models.Model):
```
erickpeirson: Hmmm, why do we need a new model for this?

annotations/views.py
```diff
@@ -364,6 +366,62 @@ def collection_texts(request, collectionid):
...
+                    if not userDict.get(username):
+                        activity = RecentActivity()
+                        activity.text +=1
+                        userDict[username] = activity
+                    else:
+                        activity = userDict.get(username)
+                        activity.text +=1
+                text_list.remove(item)
+
+    appellation_list = list(Appellation.objects.filter(created__gt=time_threshold).order_by('-created')[:10])
```
erickpeirson: Filtering by date makes sense (using `.filter(created__gt=...`) , but why then only take the ten most recent records?

annotations/views.py
apchandak24: We created mapping of user and his contributions. So to keep all his contributions at one place we made model class. Also if some more parameters are to be added in future we can add it in the model.

annotations/views.py
kardanitin: We took only 10 recent records to minimize the number of stories we need to show in recent activity. Since as per our initial discussion we needed to only consider 5 recent stories.
We can tweak the number as per our need.

erickpeirson:
@kardanitin @apchandak24 This is a really complex problem, and you guys are off to a good start. I've made several comments on lines, so take a look at that.

As far as the overall approach: if I have understood correctly, you are doing something like
1. Filtering to obtain only the most recent objects;
2. Iterating over the resulting queryset to aggregate by hour and by user

Your logic is pretty good, although the use of an extra database model is overkill (we don't need to store anything in the database). The trick now, however, is translating this (if we can) into something that can be executed at the database level rather than in Python. 

Here's a snippet to get you headed in that direction:

``` python
from django.db.models.expressions import DateTime
from django.db.models import Count
import pytz

Appellation.objects.annotate(hour=DateTime("created", "hour", pytz.timezone("UTC"))).values("hour").annotate(created_count=Count('id')).order_by("-hour")
```

Take a look, and let me know if you'd like to walk through what's going on here

annotations/views.py
kardanitin: @erickpeirson : the class won't be a database model(mistakenly inherited the class from model class ) but kindof a bean class which helps us to maintain contributions for user.

erickpeirson: @kardanitin @apchandak24 Take a look at [collections](https://docs.python.org/2/library/collections.html). We want to keep things as simple as possible, so definitely take advantage of existing classes. Even a `dict` would be enough to count beans, but something like a `Counter` or `defaultdict` (from `collections`) might be better.

But again, this may or may not be necessary after using `annotate`, as discussed.

erickpeirson: This still needs to get removed before we can merge

vogon/urls.py
erickpeirson, line 57: Looks like we still have duplicate url entries here -- lines 55 and 58

apchandak24, line 57: we have removed it and pushed the changes

annotations/views.py
apchandak24: we have removed it and pushed the changes

annotations/views.py
```diff
@@ -367,6 +371,18 @@ def collection_texts(request, collectionid):
     return HttpResponse(template.render(context))
 
 
+def recent_activity(request):
```
erickpeirson, line 374: Docstring?

kardanitin, line 374: Done

erickpeirson:
Thanks guys, great start

PR #90, opened 2016-04-14
Making preliminary changes to add a concept

yyk16293:


concepts/authorities.py
```diff
@@ -164,3 +165,13 @@ def get_by_namespace(namespace):
 
     return [ manager for manager in authority_managers
                 if manager.namespace == namespace ]
+
+def add_concept(instance):
+    """
+    Adding the approved concept to Conceptpower
+    :param instance:
```
erickpeirson: Be sure to follow the docstring style used in the rest of the project. E.g. for the `resolve()` function above:

```
Parameters
----------
sender : class
instance : :class:`.Type` or :class:`.Concept`
```

concepts/authorities.py
```diff
@@ -164,3 +165,13 @@ def get_by_namespace(namespace):
...
+    :return:
+    """
+
+    Conceptpower.create(settings.USER_ID,settings.PASSWORD,instance.label,instance.pos,
```
erickpeirson: `create()` is an instancemethod, not a static/class method.

vogon/heroku_settings.py
```diff
@@ -178,3 +178,7 @@
         'LOCATION': 'default_cache_table',
     }
 }
+
+# SuperUser Credentials
+USER_ID = 'test'
```
erickpeirson: These shouldn't be hardcoded -- see the documentation that I sent you earlier about the settings module. These values should be set in environment variables, and then loaded here via `os.environ`. Also, please use informative variable names (`USER_ID` for what?).

concepts/authorities.py
erickpeirson: Also, be sure to put a space after each comma.

erickpeirson:
Hey @yyk16293, great start! See my line comments for a few things that need to be revised.

vogon/heroku_settings.py
```diff
@@ -178,3 +178,7 @@
         'LOCATION': 'default_cache_table',
     }
 }
+
+# VOGON ADMIN USER Credentials
+VOGONADMIN_USER_ID = os.environ.get('ADMIN_ACCESS_KEY', None)
```
erickpeirson: ???? Aren't these credentials for Conceptpower

concepts/signals.py
```diff
@@ -23,6 +23,17 @@ def concept_post_save_receiver(sender, **kwargs):
     if instance:
         resolve_concept.delay(sender, instance)
 
+@receiver(post_save, sender=Concept)
```
erickpeirson, line 27: Two blank lines above all top-level class and function declarations

concepts/signals.py
```diff
@@ -3,7 +3,7 @@
 
 from .authorities import resolve
 from .models import Concept, Type
-from concepts.tasks import resolve_concept
+from concepts.tasks import resolve_concept,add_concept
```
erickpeirson: Space after comma

concepts/authorities.py
```diff
@@ -164,3 +165,16 @@ def get_by_namespace(namespace):
...
+    """
+    Adding the approved concept to Conceptpower
+
+    Parameters
+    -----------
+    instance: :class: '.Concept'
+    """
+
+    Conceptpower().create(settings.USER_ID, settings.PASSWORD, instance.label, instance.pos,
+                        'VogonWeb Concepts', instance.description, instance.typed)
```
erickpeirson: Even though we're using the same list for everything, let's declare a variable for the list name so that it's easier to document.

concepts/authorities.py
```diff
@@ -164,3 +165,16 @@ def get_by_namespace(namespace):
...
+def add(instance):
+    """
+    Adding the approved concept to Conceptpower
+
+    Parameters
+    -----------
+    instance: :class: '.Concept'
+    """
+
+    Conceptpower().create(settings.USER_ID, settings.PASSWORD, instance.label, instance.pos,
```
erickpeirson: What are these settings globals? They don't match anything in the settings modules.

vogon/mysql_settings.py
```diff
@@ -125,3 +125,6 @@
 sys.path.append('/diging/wwwnginx/vogon')
 from secrets import ERATOSTHENES_TOKEN
 ERATOSTHENES_ENDPOINT = 'http://diging.asu.edu:9000/eratosthenes'
+
+CONCEPTPOWER_USERID = os.environ.get('ADMIN_ACCESS_KEY', None)
```
erickpeirson: These global names are much better. But why would we use different names for the environment variables used to control them?

concepts/signals.py
```diff
@@ -23,6 +23,17 @@ def concept_post_save_receiver(sender, **kwargs):
...
 
+@receiver(post_save, sender=Concept)
+def concept_post_save_approve_receiver(sender, **kwargs):
+    """
+    When a :class:`.Concept` is saved, attempt to add it using one of the
+    registered :class:`.AuthorityManager` classes to ConceptPower if the
+    :class:`.Concept` is :prop:`.approved`\.
+    """
+    instance = kwargs.get('instance', None)
+    if instance and instance.concept_state == 'Approved':
```
erickpeirson: This should not be hardcoded. Use the class attribute, e.g.

``` python
if instance and instance.concept_state == Concept.APPROVED:
```

concepts/signals.py
```diff
@@ -23,6 +23,17 @@ def concept_post_save_receiver(sender, **kwargs):
...
+        add_concept.delay(sender, instance)
+
 @receiver(post_save, sender=Type)
```
erickpeirson, line 39: Two blank lines above each top-level class and function declaration

concepts/authorities.py
yyk16293: which list are you referring you to here, Erick? Can you elaborate ?

erickpeirson: `'VogonWeb Concepts',` is hardcoded

concepts/signals.py
```diff
@@ -23,6 +23,19 @@ def concept_post_save_receiver(sender, **kwargs):
...
+
+@receiver(post_save, sender=Concept)
+def concept_post_save_approve_receiver(sender, **kwargs):
+    """
+    When a :class:`.Concept` is saved, attempt to add it using one of the
+    registered :class:`.AuthorityManager` classes to ConceptPower if the
+    :class:`.Concept` is :prop:`.approved`\.
+    """
+    instance = kwargs.get('instance', None)
+    if instance and instance.concept_state == Concept.APPROVED:
```
erickpeirson, line 35: This will raise an `AttributeError` if instance is `None`

erickpeirson:
@yyk16293  This looks good  -- see my one line comment above, and then we can merge.

PR #93, opened 2016-04-15
Vgnwb 180 sprint10

ndondeti:


annotations/views.py
```diff
@@ -1040,25 +1040,58 @@ def relation_details(request, source_concept_id, target_concept_id):
...
+            'user': request.user,
+            'concept': concept,
+            'appellations': appellations_by_text,
+            'texts': texts,
+        })
+        return HttpResponse(template.render(context))
+    else:
+        response = dict()
+        response['concept'] = concept
+        json = []
```
erickpeirson, line 1066: This clobbers the `json` module, imported on line 76.

annotations/views.py
```diff
@@ -1472,7 +1505,6 @@ def generate_network_data(relationset_queryset, text_id=None, user_id=None,
     edges = {}
     nodes = {}
     seen = set([])      # Appellation ids.
-
```
erickpeirson, line 1475: Any reason why we're deleting lines down here?

annotations/views.py
```diff
@@ -1040,25 +1040,58 @@ def relation_details(request, source_concept_id, target_concept_id):
...
+            'texts': texts,
+        })
+        return HttpResponse(template.render(context))
+    else:
+        response = dict()
+        response['concept'] = concept
+        json = []
+        appellations_by_text = dict()
+        text = ""
+        for appellation in appellations:
```
erickpeirson, line 1069: You might consider using [`values()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#values) or [`values_list()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#values-list) here, for performance' sake.

PR #118, opened 2017-08-25
[VGNWB-212] fixed first character not triggering create button

tjquinn1:
When the first character is selected, this.position.startOffset = 0 which evaluates to false causing the ready function to fail. The else clause is needed because without it the range selection states true-5 instead of 0-5

### DIFF
--- a/annotations/static/annotations/js/annotators/text.js
+++ b/annotations/static/annotations/js/annotators/text.js
@@ -436,15 +436,8 @@ AppellationCreator = {
 
             // Changes value of start position to true if startOffset value is 0
             // in order to have create button show up when first character is selected
-            if(this.position.startOffset == 0) {
-                startPos = this.position.startOffset;
-                startPos = true;
-            } else {
-                startPos = this.position.startOffset;
-            }
 
-
-            return (startPos && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
+            return (this.position.startOffset >= 0 && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
         }
     }
 }
--- a/annotations/static/annotations/js/annotators/text.js
+++ b/annotations/static/annotations/js/annotators/text.js
@@ -433,11 +433,21 @@ AppellationCreator = {
             }
         },
         ready: function() {
-            return (this.position.startOffset && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
+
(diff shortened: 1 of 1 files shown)


annotations/static/annotations/js/annotators/text.js
```diff
@@ -433,12 +433,22 @@ AppellationCreator = {
             }
         },
         ready: function() {
-            return (this.position.startOffset && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
+
+            // Changes value of start position to true if startOffset value is 0
+            // in order to have create button show up when first character is selected
+            if(this.position.startOffset == 0) {
+                startPos = this.position.startOffset;
+                startPos = true;
```
jdamerow: why do you first set the value to the offset and then set it to true?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -433,12 +433,22 @@ AppellationCreator = {
             }
         },
         ready: function() {
-            return (this.position.startOffset && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
```
jdamerow, line 436: and `this.position.startOffset >= 0` didn't work?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -433,12 +433,22 @@ AppellationCreator = {
...
+            // in order to have create button show up when first character is selected
+            if(this.position.startOffset == 0) {
+                startPos = this.position.startOffset;
+                startPos = true;
+            } else {
+                startPos = this.position.startOffset;
+            }
+
+
+            return (startPos && this.position.endOffset && this.position.representation.trim().length > 0 && this.text.id && this.user.id && this.concept);
```
jdamerow: what if startOffset is > 0?

jdamerow (changes requested):

tjquinn1:
After double checking the first setting of startPos is not necessary and I
will remove it. Setting  this.position.startOffset >= 0  does not work, it
just evaluates to 0. In javascript 0 evaluates to false and in javascript
when using the and operator the function will stop evaluating after the
first false. And by startOffset > 0 do you mean if the function was written
like this.position.startOffset > 0? If not the function should runs as
normal when the value is greater than 0.

jdamerow:
yes, I mean `this.position.startOffset >= 0`. This should not evaluate to 0 but to true or false, which should be fine.

tjquinn1:
yep that works like a charm. Should I resubmit the pull request?

jdamerow (approved):

PR #121, opened 2017-09-06
Bug/vgnwb 210

tjquinn1:
Fixed filters resetting when switching pages


jdamerow:
please resolve conflicts

annotations/views/annotation_views.py
```diff
@@ -43,6 +43,15 @@ def relations(request):
...
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+
+    createdBy = request.GET.get('createdBy')
+    createdAfter = request.GET.get('createdAfter')
+    createdBefore = request.GET.get('createdBefore')
+    project = request.GET.get('project')
+    terminal_nodes = request.GET.get('terminal_nodes')
+    occursIn = request.GET.get('occursIn')
```
jdamerow: this is potentially dangerous, you can't trust that the data send in the GET request is safe. It has not been validated.

jdamerow: you should try to use the filter object instead, pass that to the template and the retrieve active filters from there.  This would also make sure the code doesn't have to be changed again if more filter fields are added.

annotations/templates/annotations/relations.html
```diff
@@ -18,7 +18,7 @@
         <div class="col-xs-4  clearfix text-center">
             {% if relations.has_previous  %}
             <div class="pull-left">
-                <a href="?page={{ relations.previous_page_number }}" class="btn btn-xs">&laquo; Previous</a>
+                <a href="?page={{ relations.previous_page_number }}{% if not createdBy %}{% else %}&createdBy={{ createdBy }}{% endif %}{% if not project %}{% else %}&project={{ project }}{% endif %}{% if not createdBefore %}{% else %}&createdBefore={{ createdBefore }}{% endif %}{% if not createdAfter %}{% else %}&createdAfter={{ createdAfter }}{% endif %}{% if not occursIn %}{% else %}&occursIn={{ occursIn }}{% endif %}{% if not terminal_nodes %}{% else %}&terminal_nodes={{ terminal_nodes }}{% endif %}" class="btn btn-xs">&laquo; Previous</a>
```
jdamerow: besides the comment below (try to use the filter object), why the empty if statement???
```
{% if not createdBy %}{% else %}&createdBy={{ createdBy }}{% endif %}
```
just unnecessary code (why not testing for the opposite with no else statement?

jdamerow (changes requested):

tjquinn1:
Copy request params. Delete page from params if present due to duplication and then encode the new params into the params context and apply when clicking link

annotations/views/annotation_views.py
```diff
@@ -43,6 +44,11 @@ def relations(request):
 
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+    gt = request.GET.copy()
+    if 'page' in gt:
+        del gt['page']
+
```
jdamerow: is there a reason you're not using the filters? This doesn't solve the problem of printing what ever is passed to the method to the page without any validation.

jdamerow (changes requested):

tjquinn1:
Fixed filters resetting by using the filter object

annotations/views/annotation_views.py
```diff
@@ -35,14 +36,22 @@ def annotate_image(request, text_id):
...
 
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+    project = filtered.data.get('project')
+    createdBy = filtered.data.get('createdBy')
+    occursIn = filtered.data.get('occursIn')
+    createdAfter = filtered.data.get('createdAfter')
+    createdBefore = filtered.data.get('createdBefore')
+    terminal_nodes = filtered.data.get('terminal_nodes')
```
jdamerow: I know you don't want to hear that, but this way your data is not yet validated. Instead, call ```is_valid()``` on the form in the filter and then put ```cleaned_data``` into your context. In your template, simply iterate over cleaned_data and append key=value to the url. This way if the filter changes, the rest of the code doesn't have to be changed accordingly -> less maintenance -> fewer possibilities for bugs.

jdamerow (changes requested):

tjquinn1:
Fixed filters by using cleaned_data. Then formatted data as needed.

annotations/views/annotation_views.py
```diff
@@ -35,14 +34,29 @@ def annotate_image(request, text_id):
...
 
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+    data = filtered.form.cleaned_data
+    dt = {}
+    for key, value in data.items():
+        if key == 'createdBy' or key  == 'project':
+            if value is not None:
+                if hasattr(value, 'id'):
```
jdamerow: those can be together in one line with and 'and'

annotations/views/annotation_views.py
```diff
@@ -35,14 +34,29 @@ def annotate_image(request, text_id):
...
     for r in qs:
         print r.__dict__
 
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+    data = filtered.form.cleaned_data
+    dt = {}
+    for key, value in data.items():
+        if key == 'createdBy' or key  == 'project':
```
jdamerow: why not ```if key in ['createdBy', 'project']:```
this would be easier to read and maintain

annotations/views/annotation_views.py
```diff
@@ -35,14 +34,29 @@ def annotate_image(request, text_id):
...
+            if value is not None:
+                if hasattr(value, 'id'):
+                    dt[key] = value.id
+        elif key == 'createdBefore' or key == 'createdAfter':
```
jdamerow: same as above

annotations/views/annotation_views.py
```diff
@@ -35,14 +34,29 @@ def annotate_image(request, text_id):
...
+    filtered = RelationSetFilter(request.GET, queryset=RelationSet.objects.all())
     qs = filtered.qs
     for r in qs:
         print r.__dict__
 
     paginator = Paginator(qs, 40)
     page = request.GET.get('page')
+
+    data = filtered.form.cleaned_data
+    dt = {}
```
jdamerow: what does 'dt' stand for? not a good variable name

jdamerow (changes requested):
almost there

jdamerow (approved):

PR #122, opened 2017-09-25
[VGNWB-209] added functionality to delete relation templates

prnvshrn:


annotations/templates/annotations/relationtemplate_list.html
```diff
@@ -26,16 +26,61 @@
...
+                          </div>
+                          <div class="modal-footer">
+                            <input class="btn btn-danger" type="submit" value="Delete">
+                            <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+                          </div>
+                          </form>
+                        </div>
+                      </div>
+                    </div>
+                    <!-- Delete Template modal section ends here -->
```
jdamerow: I wouldn't create a modal for every single template. That just becomes unmanageable when there are many templates. Rather have one modal for which you set the selected template id when the delete button is clicked.

annotations/templates/annotations/relationtemplate_list.html
```diff
@@ -46,9 +91,21 @@
...
 <script>
 jQuery(document).ready(function($) {
+
     $(".clickable-row").click(function() {
         window.document.location = $(this).data("href");
     });
+
+    $('[data-delete-template]').tooltip();
+
+    $("a[id^='openTemplateDeleteModal']").click(function(e) {
```
jdamerow: just give all buttons the same class and then use the class to identify buttons here.

annotations/views/relationtemplate_views.py
```diff
@@ -208,5 +206,12 @@ def create_from_relationtemplate(request, template_id):
     else:   # Not sure if we want to do anything for GET requests at this point.
         response_data = {}
 
-
     return JsonResponse(response_data)
+
+
+def delete_relationtemplate(request, template_id):
```
jdamerow: this needs to be access restricted.

jdamerow (changes requested):

annotations/templates/annotations/relationtemplate_list.html
prnvshrn: Changed the implementation of creating multiple modals to a single modal.

annotations/templates/annotations/relationtemplate_list.html
prnvshrn: The buttons now are identified by class.

annotations/views/relationtemplate_views.py
prnvshrn: Added "staff_member_required" before <b>delete_relationtemplate</b> method to restrict access.

prnvshrn:
I have made the following changes:-
1. Changed multiple modals to a single modal
2. Using class to identify buttons for opening the modal
3. Added 'staff_member_required' to restrict access

annotations/views/relationtemplate_views.py
```diff
@@ -208,5 +206,13 @@ def create_from_relationtemplate(request, template_id):
...
 
-
     return JsonResponse(response_data)
+
+
+@staff_member_required
+def delete_relationtemplate(request, template_id):
+    if request.method == 'POST':
+        RelationTemplate.objects.filter(id=template_id).delete()
+        RelationTemplatePart.objects.filter(id=template_id).delete()
```
jdamerow: this should be done in a transaction:
https://docs.djangoproject.com/en/1.11/topics/db/transactions/#controlling-transactions-explicitly

jdamerow (changes requested):

annotations/views/relationtemplate_views.py
prnvshrn: Changed implementation of deletion logic to execute in an atomic transaction

prnvshrn:
Changed implementation of deletion logic to execute in an atomic transaction

jdamerow:
something is going wrong here. try this: 
- create 3 templates (two simple subject-predicate-object ones, and the last one nested, not sure if this is important, but that's what I did). 
- then delete the last one (the nested one).
- no look at the first one: -> relation parts are deleted

jdamerow:
also, it looks like relationsets have references to relation templates, which means relation templates are deleted that are referenced by relationsets, there will most likely be unintended side effects. A relation template should probably not be allowed to be deleted if it is used in a relation set.

prnvshrn:
Added the following changes -
1. Wrong deletion of relation template part was occurring because in the views I was passing <b>id</b> instead of <b>part_of</b> for deletion. So I changed the method to use <b>part_of</b> instead.
2. Added check to see if data is present in relationset for a template before deletion. If data is present, an error is shown and template is not deleted.

jdamerow (approved):

PR #124, opened 2017-10-20
Bug/vgnwb 207

tjquinn1:
Filter results from goat to exclude duplicate concept identities and update template to show both identities if present


jdamerow:
please resolve conflicts

tjquinn1:
What conflicts are you getting? This runs fine on my machine

annotations/views/rest_views.py
```diff
@@ -493,8 +493,41 @@ def _relabel(datum):
...
+
+
+
+                    else:
+                        break
+
+            #print results[i]["identities"]
+            print results[i]["identities"]
+            i = i + 1
+            print results
```
prnvshrn: Please remove the extra print statements

annotations/views/rest_views.py
```diff
@@ -493,8 +493,28 @@ def _relabel(datum):
...
+        resultsLength = len(results)
+        concept1 = []
+        concept2 = []
+
+        while (i != resultsLength):
+            if results[i]["identities"]:
+                z = 1
+                while (z != len(results[i]["identities"])):
+                    concept1 = results[i]["identities"][0]["concepts"]
+                    if z != len(results[i]["identities"]):
```
hashknot, line 509: Because of the enclosing while loop, won't this condition always be true?

postdeploy.sh
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
```
hashknot, line 1: Are these changes related to the current ticket?

annotations/views/annotation_views.py
```diff
@@ -69,7 +69,6 @@ def relations(request):
     context = {
         'paginator': paginator,
         'relations': relations,
-        'params': request.GET.urlencode(),
```
hashknot, line 72: Is this change related to the current ticket?

README.md
```diff
@@ -9,138 +9,6 @@ who create those interpretations. As we build the epistemic web we are producing
...
-
-## Elasticsearch 2.x
-
-We use [Haystack](http://haystacksearch.org/) to interface with [Elasticsearch
-2](https://www.elastic.co/). The current version of Haystack doesn't actually
-support ES2 (the ES API changed dramatically), so we're using a custom backend
-located in ``annotations/elasticsearch_backends.py``.
-
-If you're using ES 1.x, the built-in ES backend for Haystack should work just
-fine.
```
hashknot, line 146: Are these changes related to the current ticket?

tjquinn1: I did not change that. my version was behind so I pulled and thats what I ended up with.

jdamerow: It seems like you merged master (instead of develop), which is why this is removed. Please revert this, as it would remove the latest changes from README.

annotations/views/rest_views.py
```diff
@@ -493,8 +493,28 @@ def _relabel(datum):
...
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
+            return {_fields.get(k, k): v for k, v in datum.iteritems() }
+        results = map(_relabel, [c.data for c in concepts])
+
+        i = 0
+        resultsLength = len(results)
+        concept1 = []
+        concept2 = []
+
+        while (i != resultsLength):
```
jdamerow, line 504: why don't you use a for loop here? you don't seem to use the 'i' for anything else than accessing the array.

annotations/views/rest_views.py
```diff
@@ -493,8 +493,28 @@ def _relabel(datum):
...
+            if results[i]["identities"]:
+                z = 1
```
jdamerow, line 506: is 'z' ever changed?

annotations/views/rest_views.py
```diff
@@ -493,8 +493,28 @@ def _relabel(datum):
...
+                while (z != len(results[i]["identities"])):
```
jdamerow, line 507: what does this while loop do?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -6,17 +6,58 @@
...
+             conceptPower :"CHPS",
+             concept2: ""
+         }
+     },
     methods: {
         select: function() {
             this.$emit('selectconcept', this.concept);
         },
 
+      ident2: function(identity2, concept, concept2) {
```
jdamerow, line 37: bad method name

annotations/static/annotations/js/annotators/text.js
```diff
@@ -6,17 +6,58 @@
...
+        if(typeof this.concept.identities[0].concepts[1] !== 'undefined'){
+          this.identity2 = true;
+          var regex = /digitalhps/g;
+          var str = this.concept.identities[0].concepts[1];
+          var m;
```
jdamerow, line 42: m, str: bad variable names

annotations/static/annotations/js/annotators/text.js
```diff
@@ -6,17 +6,58 @@
...
+
+          while ((m = regex.exec(str)) !== null) {
+              // This is necessary to avoid infinite loops with zero-width matches
+              if (m.index === regex.lastIndex) {
+                  regex.lastIndex++;
+              }
+
+              // The result can be accessed through the `m`-variable.
+              m.forEach((match, groupIndex) => {
```
jdamerow, line 51: since you don't have any groups in the pattern, shouldn't there just be one match?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -6,17 +6,58 @@
...
                    <div class="text text-muted">{{ concept.description }}</div>
-               </div>`,
+                </div>
+
+                <div v-if="identity2" class="list-group-item concept-item clearfix" id="concept-{{ concept2 }}">
+                    <div>
+                      <a v-on:click="select" style="cursor: pointer;">{{ concept.label }} ({{ conceptPower }})</a>
+
+                  </div>
+                  <div v-if="identity2" class="text text-muted">{{ concept.description }}</div>
```
jdamerow, line 22: wouldn't ```concept.label``` and ```concept.description``` have the same value as in the previous item? They should show what conceptpower actually returns.

jdamerow (changes requested):

PR #125, opened 2017-10-25
Bug/vgnwb 226

tjquinn1:
fixed scroll issue


jdamerow:
please resolve conflicts

tjquinn1:
What conflicts are you getting? This runs fine on my machine

annotations/views/repository_views.py
```diff
@@ -368,6 +368,7 @@ def repository_text_content(request, repository_id, text_id, content_id):
 
 @login_required
 def repository_text_add_to_project(request, repository_id, text_id, project_id):
+    print "yes"
```
prnvshrn: Please remove this print statement

annotations/templates/annotations/project_details.html
```diff
@@ -42,7 +42,7 @@
         <thead>
             <th class="col-xs-9">
                 <a class="h4" style="text-decoration: none; vertical-align: middle;" href="{% if order_by == '-title' %}?order_by=title{% else %}?order_by=-title{% endif %}" id="order_title">Title</a>
-                <a class="btn btn-sm btn-default" aria-label="Left Align" href="{% if order_by == '-title' %}?order_by=title{% else %}?order_by=-title{% endif %}">
+                <a class="btn btn-sm btn-default" aria-label="Left Align" href="{% if order_by == '-title' %}?order_by=title{% else %}?order_by=-title{% endif %}">\
```
prnvshrn: The \ symbol is after the anchor tag ends. Can you remove it or put it inside the anchor tag ?

annotations/templates/annotations/project_details.html
```diff
@@ -54,6 +54,7 @@
             </th>
         </thead>
         {% for text in texts %}
+        <h6>{{ text }}</h6>
```
jdamerow: this does not seem to be related to the ticket

jdamerow (changes requested):

hashknot (approved):

annotations/static/annotations/css/annotators/text.css
```diff
@@ -25,6 +25,9 @@
 
 .relation-list {
     font-size: 0.8em;
+    height: 90vh;
```
jdamerow: Ideally this should take up all the space there is, not a defined height. also there is still an issue with the footer:
image

jdamerow (changes requested):

annotations/static/annotations/css/annotators/text.css
tjquinn1: @jdamerow  I am having trouble figuring out what is wrong with this fix. I tried it on the display size for a 13in screen, ipad and ipad pro screens and all work as intended. Also vh is a percentage of the users viewport. So this element should take up 90% percent of the screen. If you want it to take up 100% of the screen.The affix property of the list/search element will have to be modified to "turn off" when a certain point is reached so that the whole thing scrolls up a little when the footer comes up otherwise the element will always stay fixed and will be in the way. This won't work on sizes smaller than an Ipad but the site is not made for mobile anyway


screen shot 2018-01-11 at 10 26 30 am
# Ipad
screen shot 2018-01-11 at 10 20 20 am

# Ipad Pro
screen shot 2018-01-11 at 10 19 46 am

# 13in Screen
screen shot 2018-01-11 at 10 19 09 am

jdamerow: hm, ja, not sure how I got the problem. Maybe I confused app versions. Let' s leave it the way it is. If it does turn out to have issues, we can always come back to it. Please reopen the PR.

jdamerow (approved):

PR #127, opened 2017-10-27
Bug/vgnwb 230

tjquinn1:
Fixed the take me back link. The link was trying to take the user to the next page instead of taking them back


prnvshrn (approved):

annotations/templates/annotations/concept_add.html
```diff
@@ -77,7 +77,7 @@
     </ul>
     {% endif %}
 
-    <a class="btn btn-md btn-danger" href="{{ next_page }}">Take me back!</a>
+    <input class="btn btn-md btn-danger" type=button value="Take me back!" onClick="javascript:history.go(-1);">
```
jdamerow: let's not use the browser history but actually fix what ever was wrong with the next_page link

jdamerow (changes requested):
also, please don't commit .DS_Store files

tjquinn1:
The problem here is that both approve and back buttons are using the next link. The next link is not broken but is instead being misused. I fixed the problem by simple having the back button return the user to the concepts page.

annotations/templates/annotations/concept_add.html
```diff
@@ -77,7 +77,7 @@
     </ul>
     {% endif %}
 
-    <a class="btn btn-md btn-danger" href="{{ next_page }}">Take me back!</a>
+    <a class="btn btn-md btn-danger" href="{% url "concepts" %}">Take me back!</a>
```
jdamerow: I believe this does not pay attention to the page the user was on? If I come to this page from the 4th page of the concepts, this will not take me back to the 4th page, will it? It should though.

jdamerow (changes requested):

tjquinn1:
Update link to return user to previous page they were on

jdamerow (approved):

PR #128, opened 2017-11-13
Bug/vgnwb 207 3

tjquinn1:
Be hold what jean and her bacon have created. It sorts duplicate concepts, it then fetches their data through xml and displays it in the template. 
 **Ticket Text**
I've used the concept before when creating appellations, and other concepts from Conceptpower are found without issues, but for some reason "Jean Bacon" is not found.
I also used the "force refresh" checkbox, but that didn't change anything.
There was an issue when I first started annotating that Conceptpower wasn't running, maybe that screw up the cache?


prnvshrn (dismissed):

hashknot (dismissed):
Discussed with Taylor offline to simplify the loop iterating on the results.

jdamerow:
@tjquinn1 did you say you would work on this more?

tjquinn1:
@jdamerow Yes

jdamerow:
can we close the PR for now then?

jdamerow:
@tjquinn1 is this done yet? if not, please close PR. If it is done, the build is failing.

tjquinn1:
@jdamerow sorry I thought I had closed this when you asked earlier, my apologies. I'm assuming it is failing due to the six thing like the other PR.

hashknot (dismissed):

annotations/views/rest_views.py
```diff
@@ -493,8 +500,76 @@ def _relabel(datum):
...
+        
+        for result in results:
+            if result["identities"]: # if identities exist append the first identitiy to the list so that we can filter out other identities against it
+                identities = result["identities"][0]["concepts"]
+            else:
+                identities = []
+            for ident in result["identities"]: # go through the identities in each result
+                for identity in identities: # go through the identities in the identities list
+                    if identity not in ident["concepts"]: # if the ideneities list does not contain the identity from the result then add it to the list
+                        identities.extend(ident["concepts"])
```
jdamerow: I'm a little confused about what line 507-514 are supposed to do. What is the goal? That you have a list with all possible URIs for that result? E.g. ["http://viaf.org/viaf/85363862", "http://www.digitalhps.org/concepts/CONpeSHC70qxNC0"] for jean bacon?
 Couldn't you just iterate over all identity objects and then all concepts and add the URIs to a set?

jdamerow: also, calling your list of concepts in line 508 'identities' is really confusing.

annotations/views/rest_views.py
```diff
@@ -493,8 +500,76 @@ def _relabel(datum):
...
+                    #list can be referenced as con0, con1, etc
+                    # This will have to be updated if a new concept Authority is added
+                    hps = re.search( r'www.digitalhps.org', concept, re.M|re.I)
+                    viaf = re.search( r'viaf.org', concept, re.M|re.I)
+                    if hps:
+                        url = concept
+                        data = urllib2.urlopen(url)
+                        tree = ET.parse(data)
+                        root = tree.getroot()
+                        namespace = {'hps': 'http://www.digitalhps.org/'}
```
jdamerow: conceptpower supports json if you'd rather use that (just add an accept header in your request).

annotations/views/rest_views.py
```diff
@@ -493,8 +500,76 @@ def _relabel(datum):
...
+                        concept_name = find(root)[0].text
+                        concept_description = find(root)[0].text
+
+                        dic = {
+                            'label': concept_name,
+                            'desc': concept_description,
+                            'uri': concept,
+                            'auth': "VIAF"}
+                        new_concepts.append(dic)
+                        i = i + 1
```
jdamerow: conceptpower and viaf parsing should be moved into their own methods.

annotations/views/rest_views.py
```diff
@@ -493,8 +500,76 @@ def _relabel(datum):
...
+                        dic = {
+                            'label': concept_name,
+                            'desc': concept_description,
+                            'uri': concept,
+                            'auth': "Concept Power"}
+                        new_concepts.append(dic)
+                        i = i + 1
+
+                    elif viaf:
+                        url = concept + '/viaf.xml' # have to append /viaf.xml to viaf url's in order to access xml
```
jdamerow: viaf also supports json if you use .json instead of .xml

annotations/views/rest_views.py
```diff
@@ -493,8 +500,76 @@ def _relabel(datum):
...
+            for ident in result["identities"]: # go through the identities in each result
+                for identity in identities: # go through the identities in the identities list
+                    if identity not in ident["concepts"]: # if the ideneities list does not contain the identity from the result then add it to the list
+                        identities.extend(ident["concepts"])
+            result["identities"] = identities # replace the identities list
+            if result["identities"]:
+                concepts = result["identities"]
+                uri = result["uri"]
+                if uri in concepts: concepts.remove(uri) # remove original uri from the list if it exists. 
+                i = 0 # used to generate concept name
```
jdamerow: maybe I'm blind, but where do you use this variable?

jdamerow (dismissed):

annotations/views/rest_views.py
```diff
@@ -493,8 +501,71 @@ def _relabel(datum):
...
 
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
+            return {_fields.get(k, k): v for k, v in datum.iteritems() }
+        results = map(_relabel, [c.data for c in concepts])
+        
+        for result in results:
+            concepts = []
+            if result["identities"]: # if identities exist append the first identitiy to the list so that we can filter out other identities against it
+                concepts = result["identities"][0]["concepts"]
```
jdamerow: you don't need this if, do you? if you start with an empty set, then first concepts list - [] will give you everything in the concept list.

annotations/views/rest_views.py
```diff
@@ -493,8 +501,71 @@ def _relabel(datum):
...
+                        concept_description = find(root)[0].text
+
+                        dic = {
+                            'label': concept_name,
+                            'desc': concept_description,
+                            'uri': concept,
+                            'auth': "VIAF"}
+                        new_concepts.append(dic)
+                    else:
+                       print "Nothing found!!"
```
jdamerow: Nope

jdamerow: totally unuseful log output

jdamerow: also, if you do want to log something, you should use a proper logger

annotations/views/rest_views.py
```diff
@@ -493,8 +501,71 @@ def _relabel(datum):
...
+                        find = e.XPath("//ns1:text", namespaces = namespace)
+                        concept_name = find(root)[0].text
+                        concept_description = find(root)[0].text
+
+                        dic = {
+                            'label': concept_name,
+                            'desc': concept_description,
+                            'uri': concept,
+                            'auth': "VIAF"}
+                        new_concepts.append(dic)
```
jdamerow: as I believe I've said before, viaf and conceptpower parsing should be done in their own methods (that are called here)

annotations/views/rest_views.py
```diff
@@ -493,8 +501,71 @@ def _relabel(datum):
...
+            if result["identities"]: # if identities exist append the first identitiy to the list so that we can filter out other identities against it
+                concepts = result["identities"][0]["concepts"]
+            conceptsSet = Set(concepts)
+            for ident in result["identities"]: # go through the identities in each result
+                identSet = Set(ident['concepts'])
+                diff = identSet.difference(conceptsSet)
+                conceptsSet.update(diff)
+            result["identities"] = list(conceptsSet) # replace the identities list
+            if result["identities"]:
+                concepts = result["identities"]
```
jdamerow: instead of this, you should be able to say
```
concepts = results["identities"] = list(conceptSet)
```

jdamerow: and then keep using ```concepts``` in the if (and if condition)

annotations/views/rest_views.py
```diff
@@ -493,8 +501,71 @@ def _relabel(datum):
...
+                uri = result["uri"]
+                if uri in concepts: concepts.remove(uri) # remove original uri from the list if it exists.
```
jdamerow: I would merge 520 and 519

requirements.txt
```diff
@@ -1,9 +1,16 @@
...
+backports-abc==0.5
+billiard==3.3.0.23
 black-goat-client==0.2.6
 bleach==1.4.2
 celery==3.1.25
+certifi==2018.1.18
+chardet==3.0.4
 conceptpower-api==1.9
+decorator==4.1.2
+dicttoxml==1.6.6
```
jdamerow: this is already imported

requirements.txt
```diff
@@ -35,8 +44,11 @@ redis==2.10.3
 requests==2.13.0
 requests-oauthlib==0.5.0
 restable>=0.1.3
+singledispatch==3.4.0.3
+six==1.10.0
+static3==0.7.0
+tornado==4.5.3
```
jdamerow: do you need all these new libraries?

jdamerow (dismissed):

annotations/views/rest_views.py
tjquinn1: Without the if it throws an index out of range error. It tries to assign `result["identities"][0]["concepts"]` to concepts that do not exist.

requirements.txt
tjquinn1: Nope I do not believe Jean Bacon required any extra libraries.

requirements.txt
tjquinn1: removed

annotations/views/rest_views.py
tjquinn1: That was just for development and I forgot to remove it.

annotations/views/rest_views.py
tjquinn1: Sorry you did.

annotations/views/rest_views.py
tjquinn1: changed

annotations/views/rest_views.py
tjquinn1: changed

annotations/views/rest_views.py
jdamerow: What I mean is that you can loose the whole if block and start with an empty concept list. the difference between an empty list and the first iteration should be everything that's in identSet (in line 550)

annotations/views/rest_views.py
```diff
@@ -493,8 +500,71 @@ def _relabel(datum):
...
+                concept_description = description1.text
+                concept_name = name1.text
+
+            dic = {
+                'label': concept_name,
+                'desc': concept_description,
+                'uri': uri,
+                'auth': "Concept Power"}
+            new_concepts.append(dic)
+            return new_concepts
```
jdamerow: you don't need to return 'new_concepts' here, do you?

annotations/views/rest_views.py
```diff
@@ -493,8 +500,71 @@ def _relabel(datum):
...
+
+        def _parseHps(self, uri, new_concepts):
+            data = urllib2.urlopen(uri)
+            tree = ET.parse(data)
+            root = tree.getroot()
+            namespace = {'hps': 'http://www.digitalhps.org/'}
+
+            for entry in root.findall('hps:conceptEntry', namespace):
+                description1 = entry.find('hps:description', namespace)
```
jdamerow: why is this called 'description1'? Is there a 'description2'?

annotations/views/rest_views.py
```diff
@@ -493,8 +500,71 @@ def _relabel(datum):
...
+            dic = {
+                'label': concept_name,
+                'desc': concept_description,
+                'uri': uri,
+                'auth': "Concept Power"}
+            new_concepts.append(dic)
+            return new_concepts
+        
+        for result in results:
+            concepts = []
```
jdamerow: do you really need this array? can't you work just with result["identities"]?

annotations/views/rest_views.py
```diff
@@ -493,8 +500,71 @@ def _relabel(datum):
...
+            if concepts:
+                if result["uri"] in concepts:
+                    concepts.remove(result["uri"]) # remove original uri from the list if it exists. 
+                new_concepts = []
+                for concept in concepts:
+                    hps = re.search( r'www.digitalhps.org', concept, re.M|re.I)
+                    viaf = re.search( r'viaf.org', concept, re.M|re.I)
+                    if hps:
+                        _parseHps(self, concept, new_concepts)
+
```
jdamerow: remove empty line

jdamerow (dismissed):

jdamerow:
Looks like there are missing imports:
NameError: global name 're' is not defined

annotations/views/rest_views.py
tjquinn1: @jdamerow If I remove concepts what will I compare results["identities"] with?

jdamerow: just initialize the set in line 545 with '[]' and further down every time you use 'concepts' use 'results["identities"]' instead. You're just giving 'results["identities"]'  a different name, don't you?

jdamerow:
cool! just caching and one box for one concept :)

tjquinn1:
@jdamerow I am not sure about the symbol and its position. I left it in the middle because it added more space between the two items and kept it from looking crowded. This symbol might go over a lot of heads if they are not math people. Maybe a equals sign or something like Bacon 1 "is related to" Bacon 2 in place of the symbol is better suited.

annotations/views/rest_views.py
```diff
@@ -477,25 +481,87 @@ def create(self, request, *args, **kwargs):
...
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
-
+        if q.lower() == 'jean bacon' and cache.get('jean bacon') != None:
+            results = cache.get('jean bacon')
+            print "success"
+        else:
```
jdamerow, line 491: ?

jdamerow (dismissed):

annotations/views/rest_views.py
tjquinn1: @jdamerow  Are you confused about the whole thing or just me leaving the print statement?

jdamerow: yes the whole thing. looks like code to check how the cache works. what would be the point in only checking for jean bacon in the cache?

tjquinn1: I only cache the jean bacon query. I check to see if jean bacon is queried and then check to make sure it exists in the cache.

jdamerow: that doesn't make much sense. why only cache one single thing if you can cache whatever is searched? just use the search string instead of the hard-coded string and it's universally applicable.

tjquinn1: I misunderstood. And just to verify this should cache only results with identities or all results?

jdamerow: just cache, what you're caching right now (all results?) and then we worry about caching concepts in general in another ticket.

annotations/views/rest_views.py
```diff
@@ -476,25 +480,87 @@ def create(self, request, *args, **kwargs):
...
-                'id': 'alt_id',
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
-
+        if cache.get(q.lower()) != None:
+            results = cache.get(q.lower())
+            print('Yes')
```
jdamerow: remove

annotations/views/rest_views.py
```diff
@@ -476,25 +480,87 @@ def create(self, request, *args, **kwargs):
...
-            _fields = {
-                'name': 'label',
-                'id': 'alt_id',
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
-
+        if cache.get(q.lower()) != None:
```
jdamerow: simply ```if cache.get(q.lower()):``` should suffice

annotations/views/rest_views.py
```diff
@@ -476,25 +480,87 @@ def create(self, request, *args, **kwargs):
...
+            results = cache.get(q.lower())
+            print('Yes')
+        else:
+            if not q:
```
jdamerow: if q is None, line 486 should already fail. this check needs to be done earlier.

annotations/views/rest_views.py
```diff
@@ -476,25 +480,87 @@ def create(self, request, *args, **kwargs):
...
+                    for concept in result["identities"]:
+                        hps = re.search( r'www.digitalhps.org', concept, re.M|re.I)
+                        viaf = re.search( r'viaf.org', concept, re.M|re.I)
+                        if hps:
+                            _parseHps(self, concept, new_concepts)
+                        elif viaf:
+                            _parseViaf(self, concept, new_concepts)
+                    result["identities"] = new_concepts # add the concept data back to the identities list
+            cache.set(q.lower(), results, 60*60*24*14) # two weeks in seconds
+            print('No')
```
jdamerow: remove

annotations/views/rest_views.py
```diff
@@ -476,25 +480,87 @@ def create(self, request, *args, **kwargs):
...
+                    new_concepts = []
+                    for concept in result["identities"]:
+                        hps = re.search( r'www.digitalhps.org', concept, re.M|re.I)
+                        viaf = re.search( r'viaf.org', concept, re.M|re.I)
+                        if hps:
+                            _parseHps(self, concept, new_concepts)
+                        elif viaf:
+                            _parseViaf(self, concept, new_concepts)
+                    result["identities"] = new_concepts # add the concept data back to the identities list
+            cache.set(q.lower(), results, 60*60*24*14) # two weeks in seconds
```
jdamerow: - that should be more like a day
- that should be a configurable value
- I just realized that if we use a cache here, the 'force refresh' checkbox needs to work and considered here (if it is checked, then the cache should not be used). Otherwise, I might search for a concept, not find it, create it in Conceptpower, but it still won't show up.

jdamerow (dismissed):

annotations/static/annotations/js/annotators/text.js
```diff
@@ -93,7 +140,7 @@ var ConceptSearch = {
                 payload['pos'] = this.pos;
             }
             if (this.force) {
-                payload['force'] = 'force';
+                payload['force'] = this.force;
```
jdamerow: shouldn't ```payload['force']``` always be what ```this.force``` is?

annotations/views/rest_views.py
```diff
@@ -476,25 +479,87 @@ def create(self, request, *args, **kwargs):
...
-        def _relabel(datum):
-            _fields = {
-                'name': 'label',
-                'id': 'alt_id',
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
+                return Response({'results': []})
```
jdamerow: is that indentation correct? seems more than the other lines.

annotations/views/rest_views.py
```diff
@@ -476,25 +479,87 @@ def create(self, request, *args, **kwargs):
...
+        force = request.GET.get('force')
 
+        if cache.get(q.lower()) and q and force == None: # If force is true we want to run the query from scratch and not use the cached results
```
jdamerow: - q needs to be the first condition or ```q.lower()``` would throw an exception if q is None
-  also: ```and not force```

annotations/views/rest_views.py
```diff
@@ -476,25 +479,87 @@ def create(self, request, *args, **kwargs):
...
-            _fields = {
-                'name': 'label',
-                'id': 'alt_id',
-                'identifier': 'uri'
-            }
-
-            return {_fields.get(k, k): v for k, v in datum.iteritems()}
-        return Response({'results': map(_relabel, [c.data for c in concepts])})
+                return Response({'results': []})
+        force = request.GET.get('force')
```
jdamerow: you probably want to default to 'False' here

annotations/views/rest_views.py
```diff
@@ -476,25 +479,87 @@ def create(self, request, *args, **kwargs):
...
+            results = map(_relabel, [c.data for c in concepts])
+
+            def _parseViaf(self, uri, new_concepts):
+                data = urllib2.urlopen(uri)
+                tree = e.parse(data)
+                root = tree.getroot()
+                namespace = {'foaf': 'http://xmlns.com/foaf/0.1/', 'ns1': 'http://viaf.org/viaf/terms#', 'owl': 'http://www.w3.org/2002/07/owl#', 'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#', 'void': 'http://rdfs.org/ns/void#'}
+                find = e.XPath("//ns1:text", namespaces = namespace)
+                concept_name = find(root)[0].text
+                concept_description = find(root)[0].text
```
jdamerow: since you don't use those variables for anything else, just put the xml expression into the dictionary creation

annotations/views/rest_views.py
```diff
@@ -476,25 +479,87 @@ def create(self, request, *args, **kwargs):
...
+                    description = entry.find('hps:description', namespace)
+                    name = entry.find('hps:lemma', namespace)
+                    concept_description = description.text
+                    concept_name = name.text
+
+                dic = {
+                    'label': concept_name,
+                    'desc': concept_description,
+                    'uri': uri,
+                    'auth': "Concept Power"}
```
jdamerow, line 535: are you only adding the last entry that has been returned into dic? what about the other ones?

jdamerow: if there is just one, you should access it like there is just one (or none)

jdamerow (dismissed):

annotations/static/annotations/js/annotators/text.js
tjquinn1: Thats what I would think. The if statement was already there so I assumed it was wanted for some reason. The false case was just handled as none. Ill remove it.

annotations/views/rest_views.py
tjquinn1: I am not sure what you mean here. I parse the xml and find the corresponding tags and then add them to the dict.

jdamerow:
please resolve conflicts

PR #130, opened 2018-01-25
Bug/vgnwb 227

tjquinn1:
Added a concept picker to choose recently and commonly used concepts instead of having to search all the time. One thing to note is I believe this exposed another bug in which the text does not unselect itself. The annotated words stay highlighted orange until the page is refreshed. The annotation list also does not work, it does not let you unselect, select annotations, hide or unhide annotations.


prnvshrn (approved):

hashknot (approved):

hashknot (approved):

jdamerow:
this will only show the most used concepts after the use has started annotating, right? It would make sense to always show the most used concepts, even if the user just opened the page (e.g. they started annotating the day before and come back to it the next day).

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,124 @@ DateAppellationCreator = {
...
+                    v-for="concept in conceptsFinal">
+                </concept-picker-item>
+               </div>`,
+    methods: {
+        selectConcept: function(concept) {
+            // Clear the concept search results.
+            this.concepts = [];
+            this.$emit('selectconcept', concept);
+        },
+        merge: function (appellations) {
```
jdamerow: are these actually appellations or concepts?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,124 @@ DateAppellationCreator = {
...
+                return b.count - a.count;
+              });
+              // filter out duplicates if concept count list and recent concept list contain the same concepts
+              for(i = 0; i <= 3; i++) {
+                if (this.conceptsFinal.includes(this.appellationsCopy[i])){
+                } else {
+                    this.conceptsFinal.push(this.appellationsCopy[i]);
+                }
+              }
+            return this.conceptsFinal;
```
jdamerow: maybe i'm missing something, but is the goal of this method to get a list of all the appellations sorted by count? if so, this looks a little too complicated. Why don't you create maps of the form {uri: count} and {uri: appellation} and then use the count map to create a list out of the appellations map?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,124 @@ DateAppellationCreator = {
...
+                }
+            }
+            // sort appellations copy by count
+            this.appellationsCopy.sort(function (a, b) {
+                return b.count - a.count;
+              });
+              // filter out duplicates if concept count list and recent concept list contain the same concepts
+              for(i = 0; i <= 3; i++) {
+                if (this.conceptsFinal.includes(this.appellationsCopy[i])){
+                } else {
```
jdamerow: empty if statements are bad coding; just makes it harder to read and maintain. if you have an empty if statement, you just need to turn around the condition to move the else content into the if block.

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,124 @@ DateAppellationCreator = {
...
+                // find index of count in array.
+                if (i === 0) {
+                    idx = 2;
+                } else { 
+                    idx = (idx + 3);
+                }
+                this.appellationsCopy[i]["count"] = this.concepts[idx];
+                if (i <= 3) { // add only the first 4 concepts to the list
+                    this.conceptsFinal.push(this.appellationsCopy[i]);
+                }
```
jdamerow: why is this here? Don't line 417-421 the same thing?

jdamerow (dismissed):

tjquinn1:
@jdamerow  As long as you have started making annotations it will display the most used concepts. If you highlight something it will insert the concept picker, if you just load the page and do not highlight anything you will not see it.

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,105 @@ DateAppellationCreator = {
...
+    methods: {
+        selectConcept: function(concept) {
+            // Clear the concept search results.
+            this.concepts = [];
+            this.$emit('selectconcept', concept);
+        },
+        merge: function (appellations) {
+            this.conceptsFinal = [];
+            this.appell = appellations;
+            this.conceptsFinal = this.appell.slice(0,4)
```
jdamerow: why is this initialized with the first 4 entries?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,105 @@ DateAppellationCreator = {
...
+            
+            this.appellationMap = this.appellations.map(function (concept, index, array) {
+                return concept.interpretation.uri; 
+                
+            });
+            var apellationCount = _.countBy(this.appellationMap); //lodash
```
jdamerow: appellation is misspelled

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,105 @@ DateAppellationCreator = {
...
+
+              this.appell.sort(function (a, b) {
+                return b.count - a.count;
+              });
+              
+              for(i = 0; i <= 4; i++) {
+                if (!this.conceptsFinal.includes(this.appell[i])){
+                    this.conceptsFinal.push(this.appell[i]);
+                } 
+              }
```
jdamerow: this will not add 4 entries if there are duplicates. Is that intended

jdamerow: also, what if there are less than 4 appellations?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,105 @@ DateAppellationCreator = {
...
+            // Clear the concept search results.
+            this.concepts = [];
+            this.$emit('selectconcept', concept);
+        },
+        merge: function (appellations) {
+            this.conceptsFinal = [];
+            this.appell = appellations;
+            this.conceptsFinal = this.appell.slice(0,4)
+            
+            this.appellationMap = this.appellations.map(function (concept, index, array) {
```
jdamerow: appellationMap is not actually a map, is it? It's just an array? It shouldn't be called Map if it isn't one.

jdamerow (dismissed):

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,119 @@ DateAppellationCreator = {
...
+                if (!this.conceptsFinal.includes(this.appell[i])){
+                    this.conceptsFinal.push(this.appell[i]);
+                    i++;
+                    count++;
+                } else {
+                    i++;
+                }
+            }
+            return this.conceptsFinal;
+        },
```
jdamerow: I'm very confused by this. let's talk about it on Thursday or Friday.

tjquinn1 (commented):
Updated

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+            }
+              
+            this.appell.sort(compare);
+            var appellLength = this.appell.length
+            // set end slice length incase it is less than 4 slice the total length of array.
+            if (appellLength < 4) {
+                sliceAt = appellLength;
+            } else {
+                sliceAt = 4
+            }
```
jdamerow: replace with:
```
var sliceAt = appellLength < 4 ? appellLength : 4;
```
also, sliceAt is not declare anywhere is it? and 4 should be a constant to make it easily changable.

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+            this.conceptsFinal = this.appell.slice(0,sliceAt);
```
jdamerow: needs a comment why you do this

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+            var appellationMap = new Map();
+            // group appellations to get total count
+            this.appell.forEach(function(item){ 
+                if (appellationMap.has(item.interpretation.uri)) {
+                   appellationMap.get(item.interpretation.uri).push(item);
+                    } else {
+                        appellationMap.set(item.interpretation.uri, [item]);
+                    }
```
jdamerow: formatting off

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+              });
+            // sort appellationMap by length
```
jdamerow: same here, needs a comment why you do that

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+            var sortedMap = new Map([...appellationMap.entries()].sort(function (a, b) {
+                return b[1].length - a[1].length
+            }));
+            // add none duplicate appellations to conceptsFinal, up to 4 total.
+            // If the iterator runs out of values before the count is filled break out of loop
+            var sortedMapItem = sortedMap.entries();
```
jdamerow: these are several entries, right? so should be 'Items' not 'Item'?

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +329,129 @@ DateAppellationCreator = {
...
+                if (appellation != null) {
+                    console.log(appellation[1][0]);
+                    if (!this.conceptsFinal.includes(appellation[1][0])){
+                        this.conceptsFinal.push(appellation[1][0]);
+                        count++;
+                    }
+                } else {
+                    break
+                }
+            }
```
jdamerow: several things:
-  you should be able to just do ```while(var count ...```
- what if sortedMap has less than 3 items? there will be an exception thrown in line 414 (undefined has no property value)
- remove log message
- what does ```appellation[1][0]``` refer to? shouldn't that simply take the first element from the list, since appellation is the value of the entry => list of appellations?

jdamerow (dismissed):

annotations/static/annotations/js/annotators/text.js
```diff
@@ -327,24 +330,123 @@ DateAppellationCreator = {
...
+                if (appellation == null) {
+                    break
+                } 
+                if (!this.conceptsFinal.includes(appellation[1][0])){
+                    this.conceptsFinal.push(appellation[1][0]);
+                    this.count++;
+                }
+            }
+            // reset count
+            this.count = 0;
```
jdamerow: is there a reason that this is not a local variable?

tjquinn1: Nope, I guess I just enjoy putting 'this' in front of everything.

jdamerow: 😆

jdamerow:
looks good! :)

PR #131, opened 2018-02-01
Bug/vgnwb 237

tjquinn1:
I added code to ignore unicode errors so that text will still display regardless if there are errors. The formatting is weird if there are errors because the unknown characters are stripped. From what I have seen the only way to solve this problem would to save the text codecs when uploading a text then using that to decode the text when displaying it.

### DIFF
--- a/annotations/annotators.py
+++ b/annotations/annotators.py
@@ -160,7 +160,7 @@ class Annotator(object):
             'text': self.text,
             'textid': self.text.id,
             'title': 'Annotate Text',
-            'content': content.decode(detect['encoding']).encode('utf-8'), 
+            'content': content.decode(detect['encoding']).encode('utf-8'), # We are using chardet to guess the encoding becuase giles is returning everyting with a utf-8 header even if it is not utf-8
             'baselocation' : basepath(request),
             'userid': request.user.id,
             'title': self.text.title,
--- a/annotations/annotators.py
+++ b/annotations/annotators.py
@@ -153,11 +153,12 @@ class Annotator(object):
     def get_context(self):
         resource = self.get_resource()
         request = self.context.get('request')
+        content = self.get_content(resource)
         return {
             'text': self.text,
             'textid': self.text.id,
             'title': 'Annotate Text',
-            'content': self.get_content(resource),
+            'content': unicode(content, errors='ignore'),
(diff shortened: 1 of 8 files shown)


annotations/views/rest_views.py
```diff
@@ -493,8 +500,79 @@ def _relabel(datum):
...
+                    if set(identity) != set(ident["concepts"]): # if the ideneities list does not contain the identity from the result then add it to the list
+                        identities.append(ident)
+            result["identities"] = identities # replace the identities list
+            if result["identities"]:
+                concepts = result["identities"]
+                uri = result["uri"]
+                if uri in concepts[0]: concepts[0].remove(uri) # remove original uri from the list if it exists. 
+                i = 0 # used to generate concept name
+                new_concepts = {}
+                for concept in concepts[0]: # determine if the concept is a viaf or concept power uri
```
hashknot: What about concepts with indices greater than 0?

hashknot (approved):

prnvshrn (approved):

jdamerow:
The character encoding is actually provided the in the header in the content-type. There is a clash between the content type the server serves and what the user uploads, but vogon should probably still get the encoding from the header in case we fix the encoding clash in giles.

tjquinn1:
Not sure if this is what you were looking for but this code should be able to encode all text to utf-8 regardless what encoding is stored in giles. I have checked with a couple files that use Windows-1252 and it does display everything correctly.

annotations/annotators.py
```diff
@@ -153,11 +154,13 @@ def render_display(self, context={}):
...
         resource = self.get_resource()
         request = self.context.get('request')
+        content = self.get_content(resource)
+        detect  = chardet.detect(content)
         return {
             'text': self.text,
             'textid': self.text.id,
             'title': 'Annotate Text',
-            'content': self.get_content(resource),
+            'content': content.decode(detect['encoding']).encode('utf-8'),
```
jdamerow: let's put a comment here that says we need to do this guessing because of a bug in Giles (it returns everything with a utf-8 header, even if the text is not in utf-8)

tjquinn1: @jdamerow  Done

PR #132, opened 2018-02-27
Story/vgnwb 233

tjquinn1:
Relationships should not be submitted to Quadriga for a certain period of time to allow users to correct mistakes.

This should be a configurable value. Let's make it 24 hours for now.


annotations/tasks.py
```diff
@@ -235,16 +237,20 @@ def accession_ready_relationsets():
...
-                def _state(obj):
-                    obj.pending = True
-                    obj.save()
-                map(_state, user_rsets)
-                submit_relationsets_to_quadriga.delay(map(lambda o: o.id, user_rsets), text_id, user_id, **kwargs)
+            timeCreated = relationset.created
+            timeCheck = datetime.now(timezone.utc)
+            timeDifference = timeCheck - timeCreated
+            result = divmod(timeDifference.days * 86400 + timeDifference.seconds, 60)
+            if result[0] >= settings.SUBMIT_WAIT_TIME:
```
jdamerow: wouldn't it be easier and prettier to use timedelta and say: 
```if timeCreated + timedelta(wait time) > now```

**Review by jdamerow — DISMISSED** [review]

**Review by prnvshrn — APPROVED** [review]

**Comment by jdamerow** [comment]
please resolve conflicts

File: vogon/settings.py [hunk 2]
```diff
@@ -257,3 +257,9 @@
     'viaf:corporate': CORPORATE_CONCEPT_TYPE,   # E40 Legal Body
     'viaf:geographic': GEOGRAPHIC_CONCEPT_TYPE,    # E53 Place
 }
+
+SUBMIT_WAIT_TIME = {
+        'days': 24,
```
jdamerow: let's make that 3 days rather than 24. 24 seems pretty long.

jdamerow (changes requested):
seems like there is a missing requirement:
ImportError: No module named chardet

I guess I should have run your previous PR locally before merging it...  ;-p

jdamerow:
It s

vogon/settings.py
tjquinn1: @jdamerow  Hahaha whoops, I meant 24 hours but Ill change it to 3 days

requirements.txt
```diff
@@ -34,10 +43,12 @@ pytz==2016.10
...
-restable>=0.1.3
+requirements==0.1
+restable==0.1.3
+singledispatch==3.4.0.3
+six==1.10.0
+static3==0.7.0
+tornado==4.5.3
 Unidecode==0.4.20
 urllib3==1.19.1
 xmltodict==0.10.2
```
jdamerow: are all these additional requirements necessary?

jdamerow (changes requested):

requirements.txt
tjquinn1: Man that was fast. The only thing I installed in this ticket was chardet do I assumed that these all belonged with it. I looked at the documentation and didn't see anything about the dependencies so I have removed what I think is unnecessary.

PR #133, opened 2018-08-17
Story/vgnwb 251

tjquinn1:
Upload the csv from the jupyter notebook and get text is necessary.  This works but could use some improvment

### DIFF
--- a/annotations/templates/annotations/project_details.html
+++ b/annotations/templates/annotations/project_details.html
@@ -24,6 +24,9 @@
                     href="{% url "edit_project" project.id %}">
                     <span class="gylphicon glyphicon-pencil"></span> Edit Project
                 </a>
+                <a class="btn btn-sm btn-info" href="{% url 'upload' %}">
+                <span class="gylphicon glyphicon-pencil"></span> Upload
+                </a>
             </span>
             {% endif %}
         </div>
@@ -99,5 +102,4 @@
     {% include "annotations/relations_list_fragment.html" %}
     {% endif %}
 </div>
-
 {% endblock %}
--- a/annotations/templates/annotations/project_details.html
+++ b/annotations/templates/annotations/project_details.html
@@ -24,8 +24,8 @@
                     href="{% url "edit_project" project.id %}">
                     <span class="gylphicon glyphicon-pencil"></span> Edit Project
                 </a>
-                <a class="btn btn-sm btn-info" href="{% url 'upload' %}">
-                <span class="gylphicon glyphicon-pencil"></span> Upload
+                <a class="btn btn-sm btn-info" href="{% url 'import' project_id=project.id %}">
+                <span class="gylphicon glyphicon-pencil"></span> Import Appellations
                 </a>
             </span>
             {% endif %}
--- a/annotations/views/project_views.py
+++ b/annotations/views/project_views.py
@@ -15,19 +15,12 @@ from django.core.files.storage import FileSystemStorage
 from annotations.models import TextCollection, RelationSet, Appellation, Text, DocumentPosition
 from annotations.forms import ProjectForm, ImportForm
 from concepts.models import Concept
-import io
 import requests
 import unicodecsv as csv_unicode
-import csv
 from repository_views import repository_text_content
 from repository import auth
 from repository.models import Repository
 from repository.managers import RepositoryManager
-import tempfile
-import os
-from django.core.files.storage import default_storage
-from django.core.files.base import ContentFile
-import datetime
 from annotations.tasks import process_import_task
 
 def view_project(request, project_id):
--- a/annotations/views/project_views.py
+++ b/annotations/views/project_views.py
@@ -13,16 +13,22 @@ from django.conf import settings
 from django.db.models import Q, Count
 from django.core.files.storage import FileSystemStorage
 from annotations.models import TextCollection, RelationSet, Appellation, Text, DocumentPosition
-from annotations.forms import ProjectForm, PathForm
+from annotations.forms import ProjectForm, ImportForm
 from concepts.models import Concept
-import unicodecsv as csv
 import io
 import requests
+import unicodecsv as csv_unicode
+import csv
 from repository_views import repository_text_content
 from repository import auth
 from repository.models import Repository
 from repository.managers import RepositoryManager
 import tempfile
+import os
+from django.core.files.storage import default_storage
+from django.core.files.base import ContentFile
+import datetime
+from annotations.tasks import process_import_task
 
 def view_project(request, project_id):
     """
@@ -189,11 +195,7 @@ def list_projects(request):
     }
     return render(request, template, context)
 
-
-@login_required
-def import_appellation(request, project_id):
-    print("TEST")
-    def add_text_to_project(request, repository_id, text_id, project_id):
+def add_text_to_project(request, repository_id, text_id, project_id):
         repository = get_object_or_404(Repository, pk=repository_id)
         project = get_object_or_404(TextCollection, pk=project_id)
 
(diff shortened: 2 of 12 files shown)


annotations/templates/annotations/project_details.html
```diff
@@ -24,6 +24,9 @@
                     href="{% url "edit_project" project.id %}">
                     <span class="gylphicon glyphicon-pencil"></span> Edit Project
                 </a>
+                <a class="btn btn-sm btn-info" href="{% url 'upload' %}">
+                <span class="gylphicon glyphicon-pencil"></span> Upload
```
jdamerow: Is this for importing appellations? If so, it should say 'Import Appellations'.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
     return render(request, template, context)
+
+
+
+def upload(request):
+
+    context = {}
+    pathForm = PathForm()
+    
+    context = {}
```
jdamerow: already defined

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                position_id = pos.id
+                )
+        else:
+            context ={
+                'form': pathForm
+            }
+    else:
+        context ={
+                'form': PathForm()
+            }
```
jdamerow: just initiate the context with the form in the beginning and loose the else block.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+
+
+def upload(request):
+
+    context = {}
+    pathForm = PathForm()
+    
+    context = {}
+    if request.method == 'POST':
+        pathForm = PathForm(request.POST)
```
jdamerow: what does 'PathForm' mean? What is the form for? Needs a better name.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                Appellation.objects.create(
+                stringRep = row[0],
+                startPos = row[1],
+                endPos = row[2],
+                createdBy = request.user,
+                occursIn = occur,
+                project_id = project,
+                interpretation = Concept.objects.get(id=row[4]),
+                position_id = pos.id
+                )
```
jdamerow: all of this should not be in the view. Ideally this should be a task, since depending on the file size this might take quite a while.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+    context = {}
+    if request.method == 'POST':
+        pathForm = PathForm(request.POST)
+        if pathForm.is_valid():
+            csv_file = request.FILES['csv_file']
+            decoded_file = csv_file.read().decode('utf-8')
+            io_string = io.StringIO(decoded_file)
+            csvreader = csv.reader(io_string, delimiter=',', quotechar='"')
+            csvreader.next()
+            project = request.session.get('project', 'none')
```
jdamerow: the project id should come from the request, not the session.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+
+    context = {}
+    pathForm = PathForm()
+    
+    context = {}
+    if request.method == 'POST':
+        pathForm = PathForm(request.POST)
+        if pathForm.is_valid():
+            csv_file = request.FILES['csv_file']
+            decoded_file = csv_file.read().decode('utf-8')
```
jdamerow: just use 'unicodecsv' and open it directly as utf-8

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+            io_string = io.StringIO(decoded_file)
+            csvreader = csv.reader(io_string, delimiter=',', quotechar='"')
+            csvreader.next()
```
jdamerow: is that to skip the headers? just use 'DictReader' and read cells by providing their headers. It's more readable and less error prone (also column order doesn't matter if you do it like that).

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+            project = request.session.get('project', 'none')
+            for row in csvreader:
+                try:
+                    parent = Text.objects.get(uri=row[3])
+                    text = Text.objects.get(part_of_id=parent.id)
+                    occur = text
+                except:
```
jdamerow: what exception are you catching here?

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+            csv_file = request.FILES['csv_file']
+            decoded_file = csv_file.read().decode('utf-8')
+            io_string = io.StringIO(decoded_file)
+            csvreader = csv.reader(io_string, delimiter=',', quotechar='"')
+            csvreader.next()
+            project = request.session.get('project', 'none')
+            for row in csvreader:
+                try:
+                    parent = Text.objects.get(uri=row[3])
+                    text = Text.objects.get(part_of_id=parent.id)
```
jdamerow: can there be only one child text object? that doesn't sound right.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                    occur = text
+                except:
+                    url = "https://amphora.asu.edu/amphora/resource/get?uri=" + row[3] + "&format=json"
```
jdamerow: that should not be a hard coded string.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                    r = requests.get(url, headers=auth.jars_github_auth(request.user))
```
jdamerow: bad variable name

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                    text_json = r.json()
+                    found = False
+                    while found == False:
```
jdamerow:
```
while not found
```

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                        for content in text_json['content']:
+                            if content['content_resource']['content_type'] == 'text/plain':
+                                text_content = content['content_resource']['id']
+                                found = True
```
jdamerow: why do you have a double loop here? Also, this looks dangerously like a possible endless loop (what if there is no plain text content type)?

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                    repository_text_add_to_project(request, 1, text_json['id'],project)
+                    text = Text.objects.get(uri=row[3])
+                    repository_text_content(request, 1, text_json['id'], text_content)
```
jdamerow: don't call a view here. especially since you're not using the returned http object. If the view does something that you need, put that something into it's own method and call it from both views.

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
         'projects': qs,
     }
     return render(request, template, context)
+
+
+
+def upload(request):
```
jdamerow: only owner should be allowed to do this.

jdamerow: also, bad view name. this is for importing appellations, not for uploading (although it does contain an upload, this is not it's main purpose).

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+
+                Appellation.objects.create(
+                stringRep = row[0],
+                startPos = row[1],
+                endPos = row[2],
+                createdBy = request.user,
+                occursIn = occur,
+                project_id = project,
+                interpretation = Concept.objects.get(id=row[4]),
+                position_id = pos.id
```
jdamerow: indentation

annotations/views/project_views.py
```diff
@@ -179,3 +185,67 @@ def list_projects(request):
...
+                )   
+
+                Appellation.objects.create(
+                stringRep = row[0],
+                startPos = row[1],
+                endPos = row[2],
+                createdBy = request.user,
+                occursIn = occur,
+                project_id = project,
+                interpretation = Concept.objects.get(id=row[4]),
```
jdamerow: what if there is no concept with the id?

vogon/urls.py
```diff
@@ -91,6 +91,7 @@
     url(r'^project/(?P<project_id>[0-9]+)/edit/$', views.project_views.edit_project, name='edit_project'),
     url(r'^project/create/$', views.project_views.create_project, name='create_project'),
     url(r'^project/$', views.project_views.list_projects, name='list_projects'),
+    url(r'^project/upload/$', views.project_views.upload, name='upload'),
```
jdamerow: this should have the project id int he url (like edit project) and be called 'import' instead of 'upload'

jdamerow (dismissed):

annotations/views/project_views.py
tjquinn1: Why send the form again if it is not needed?

annotations/templates/annotations/project_details.html
tjquinn1: changed

annotations/views/project_views.py
tjquinn1: removed

annotations/views/project_views.py
jdamerow: if you don't want to set it again, just reverse the if, but this is really hard to read (really long if block and then two else's with almost the same code one right after the other.

annotations/views/project_views.py
tjquinn1: removed

annotations/views/project_views.py
tjquinn1: i removed the form

annotations/views/project_views.py
tjquinn1: changed and fixed

annotations/views/project_views.py
tjquinn1: removed

annotations/views/project_views.py
tjquinn1: removed

annotations/views/project_views.py
tjquinn1: Changed both

annotations/views/project_views.py
tjquinn1: removed

annotations/views/project_views.py
tjquinn1: Changed

vogon/urls.py
tjquinn1: changed

annotations/views/project_views.py
tjquinn1: changed

annotations/views/project_views.py
tjquinn1: Changed

annotations/views/project_views.py
tjquinn1: moved to settings

annotations/views/project_views.py
tjquinn1: changed

annotations/views/project_views.py
tjquinn1: changed

annotations/views/project_views.py
tjquinn1: fixed

tjquinn1:
@jdamerow 
A couple things:

1.  I did not use dictreader since unicode csv returns a list of lists and that does not work with dictreader.
2. I did not go with the temp files since celery pickles what ever is passed into the task. Maybe I misunderstood how pickling works but it does not make since to make the temp file and then pickle it, might as well just pickle the data and skip the temp file.
3. From here it seemed that you only wanted the creation of the appellations in the task but maybe we should have the whole thing take place in a task since it could take a while to go through the spreadsheet and create the text.
4. Lastly I was not sure about showing the user the state of the task. I know that we can check the state but wasn't sure if there should be another task to check the state or uses signals or something like that.

jdamerow:
1. the way it is right now looks fine
2. Not a 100% sure what you were trying to do, but you wouldn't have to pickle the file. You would store the uploaded file as a temp file. then pass the path to the file into the task. the task picks up the path, opens the file, voila! Passing all the data into the task isn't a very clear approach. What if the file is huge? It'll load all of that into memory, which is unnecessary.
3. ya, it makes sense to put the whole thing into a task.
4. you would have a task object (create a new class for tasks) and store that in the database. your task would have several steps, so your task class would have a field that counts the steps, and a field for the total nubmer of steps (e.g. 50 rows + 1 adding a file = 51 steps, and after every row you work on you update the done steps by 1, store the object back into the db). in the gui, you can read out the object from the db and show how far the task has gotten (e.g. 30 out of 51).

PR #134, opened 2018-08-28
[VGNWB-256] fixed project not opening when an empty text is included …

tjquinn1:
Fixed empty texts from prevent collections from opening by removing empty text.

### DIFF
--- a/annotations/views/repository_views.py
+++ b/annotations/views/repository_views.py
@@ -97,13 +97,12 @@ def repository_collection(request, repository_id, collection_id):
         collection = manager.collection(id=collection_id, **params)
     except IOError:
         return render(request, 'annotations/repository_ioerror.html', {}, status=500)
-
     project_id = request.GET.get('project_id')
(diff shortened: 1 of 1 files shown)


annotations/views/repository_views.py
```diff
@@ -97,21 +97,20 @@ def repository_collection(request, repository_id, collection_id):
...
+    resources = collection.get('resources', [])
     context = {
         'user': request.user,
         'repository': repository,
         'collection': collection,
         'collection_id': collection_id,
         'title': 'Browse collections in %s' % repository.name,
         'project_id': project_id,
-        'resources': collection.get('resources', []),
+        'resources': [resource for resource in resources if resource['url'] is not None],
```
jdamerow: you can probably just say 
```if resource['url']```
```is not None``` is not necessary I believe.

jdamerow (dismissed):

prnvshrn (approved):

PR #135, opened 2018-11-20
Story/vgnwb 251

tjquinn1:
Allow file import and create progress bar


prnvshrn (approved):

annotations/migrations/0034_auto_20181113_1716.py
```diff
@@ -0,0 +1,27 @@
...
+            field=models.IntegerField(default=0),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name='importtasks',
+            name='total_rows',
+            field=models.IntegerField(default=0),
+            preserve_default=False,
+        ),
+    ]
```
jdamerow: let's make those two migrations one, since they are both for the same class.

annotations/migrations/0038_remove_importtasks_task_id.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('annotations', '0037_auto_20181113_1945'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='importtasks',
+            name='task_id',
+        ),
+    ]
```
jdamerow: actually, make all of these migrations one.

annotations/models.py
```diff
@@ -1143,3 +1142,14 @@ class DocumentPosition(models.Model):
...
     blank.
     """
+
+class ImportTasks(models.Model):
+    """
+    Records the event that a set of :class:`.RelationSet`\s are accessioned to
+    Quadriga.
+    """
+    created = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey('VogonUser')
```
jdamerow: Why do you use a string here and not the class?

annotations/models.py
```diff
@@ -1143,3 +1142,14 @@ class DocumentPosition(models.Model):
...
+    file_name = models.CharField(max_length=255, blank=False, null=False)
+    total_rows = models.IntegerField(blank=True, null=True)
+    c_row = models.IntegerField(blank=True, null=True)
```
jdamerow: bad variable name. what does this mean?

annotations/models.py
```diff
@@ -1143,3 +1142,14 @@ class DocumentPosition(models.Model):
...
     """
+
+class ImportTasks(models.Model):
+    """
+    Records the event that a set of :class:`.RelationSet`\s are accessioned to
+    Quadriga.
+    """
+    created = models.DateTimeField(auto_now_add=True)
+    user = models.ForeignKey('VogonUser')
+    file_name = models.CharField(max_length=255, blank=False, null=False)
```
jdamerow: there are filenames that are longer than that... (not sure why people do that but it happens)

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
     if file_content != None:
         tokenized_content = tokenize(file_content)
         return save_text_instance(tokenized_content, text_title, date_created, is_public, user, uri)
+
+
+
+
+
+
```
jdamerow: why so many empty lines?

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+@shared_task
+def process_import_task(user, project_id ,file, part_of_id, action, name):
+    with open(file, 'r') as f:
```
jdamerow: put a try catch around this. sometimes files can't be open or something else happens and celery just silently failed. you still want to update the task though, so that the user knows that something failed.

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+        csv_file = csv.reader(f)
+        # there isn't a good way to reset the iterator so we tee them up
+        reader1, reader2 = tee(csv_file)
+        reader1.next()
+        reader2.next()
+        total_count = len(list(reader2))
```
jdamerow: presumably, this will load the whole csv file into memory, won't it? that something you want to avoid, since it can cause issues with large files.

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+
+
+
+
+@shared_task
+def process_import_task(user, project_id ,file, part_of_id, action, name):
+    with open(file, 'r') as f:
+        csv_file = csv.reader(f)
+        # there isn't a good way to reset the iterator so we tee them up
+        reader1, reader2 = tee(csv_file)
```
jdamerow: you should be able to set the filereader to its beginning to reset the iterator: f.seek(0) and then recreate the reader. this works probably too though. just name them better.

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+        task = ImportTasks.objects.create(
+            user = user,
+            file_name = name,
+            total_rows = int(total_count)
+        )
+        for row in reader1:
+            # Store url so we can skip making the request to fetch the json if the urls are the same
+            stored_url = ''
+            try:
+                parent = Text.objects.get(uri=row[3])
```
jdamerow: let's use the dictreader. This will make the code much more readable.

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+                text = Text.objects.get(part_of_id=parent.id)
+                occur = text
+            except:
+                # Only make the request and get the json if the url has changed.
+                url = settings.AMPHORA_RESOURCE_URL + row[3] + "&format=json"
+                if stored_url != url:
+                    stored_url = url
+                    text_request = requests.get(url, headers=auth.jars_github_auth(user))
+                    print(auth.jars_github_auth(user))
```
jdamerow: remove

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+                    text_json = text_request.json()
+                    #find the first text in content
+                    for content in text_json['content']:
+                        if content['content_resource']['content_type'] == 'text/plain':
+                            text_content = content['content_resource']['id']
+                            # Save time and stop the iteration
+                            break
```
jdamerow: downloading a text from amphora should be in its own method (preferably in its own package)

annotations/tasks.py
```diff
@@ -278,3 +293,69 @@ def handle_file_upload(request, form):
...
+                createdBy = user,
+                occursIn = occur,
+                project_id = project_id,
+                interpretation = Concept.objects.get(id=row[4]),
+                position_id = pos.id
+            )
+            count = count + 1
+            task.c_row = int(count)
+            task.save()
+    default_storage.delete(file)
```
jdamerow, line 346: see my above comment about try/catch. you probably always want to delete the uploaded file, even if there was an exception somewhere.

annotations/templates/annotations/appellation_upload.html
```diff
@@ -0,0 +1,27 @@
...
+{% load staticfiles %}
+
+{% block extrahead %}
+<script src="{% static "annotations/js/spin.min.js" %}"></script>
+<script src="{% static "annotations/js/jquery.spin.js" %}"></script>
+<script src="{% static "annotations/js/cytoscape.min.js" %}"></script>
+{% endblock %}
+
+{% block main %}
+<div class="form-group">
```
jdamerow: let's tell the user somewhere here, what they are about to do.

annotations/text_upload.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+
+
+def repository_text_content(user, repository_id, text_id, content_id, part_of_id, action, project_id):
+    repository = get_object_or_404(Repository, pk=repository_id)
+
+    manager = RepositoryManager(repository.configuration, user=user)
+    # content_resources = {o['id']: o for o in resource['content']}
+    # content = content_resources.get(int(content_id))    # Not a dict.
```
jdamerow: ?

annotations/text_upload.py
```diff
@@ -0,0 +1,94 @@
...
+    try:
+        content = manager.content(id=int(content_id))
+        resource = manager.resource(id=int(text_id))
+    except IOError:
+        return HttpResponse("IO Error", content_type="text/plain")
```
jdamerow: why do you return an httpresponse object here?

jdamerow: wait, where is this called? is this a view? i'm confused.

annotations/text_upload.py
```diff
@@ -0,0 +1,94 @@
...
+
+    content_type = content.get('content_type', None)
+    from annotations import annotators
```
jdamerow: why is that imported here?

annotations/text_upload.py
```diff
@@ -0,0 +1,94 @@
+
+from annotations.models import Text, TextCollection, RelationSet
```
jdamerow, line 2: this file needs documentation

annotations/views/project_views.py
```diff
@@ -179,3 +195,35 @@ def list_projects(request):
...
+    text_collection = TextCollection.objects.get(id = project_id)
+    if request.method == 'POST' and request.user == text_collection.ownedBy:
+        data = request.FILES['csv_file']
+        # read csv file
+        csv_file = csv_unicode.reader(data)
+        # Create Temp File.
+        name = 'tmp/'+ request.user.username + '_' + str(datetime.datetime.now())+'.csv'
+        tmp_file = default_storage.save(name, ContentFile(data.read()))
+        # write csv contents to temp file
+        with open(tmp_file, 'w') as f:
```
jdamerow: why do you do that? doesn't this ```default_storage.save(name, ContentFile(data.read()))``` already do that, or why do you give it a contentfile?

annotations/views/project_views.py
```diff
@@ -179,3 +195,35 @@ def list_projects(request):
...
+            writer = csv.writer(f)
+            for row in csv_file:
+                try:
+                    writer.writerow(row)
+                except Exception as e:
+                    print ('Error in writing row:',e)
```
jdamerow: presumably, if one row can't be written, all of them can't. or at least it'll result in an incomplete file. Either way, the user should know about it.

jdamerow (dismissed):

annotations/text_upload.py
tjquinn1: I didn't write this. This started life as a view which basically does everything I need it to do so I repurposed it into a helper function. I did not spend time to throughly figure out how it works it magic but i use it to add the text to the users repo if it does not exist already. I am not sure why that it commented out but thats the way I found it. Ill remove the extra lines.

annotations/text_upload.py
tjquinn1: Just like the above comment it started life as a view and has since moved onto greater things in the form of a helper function. I missed the httpResponses, I will fix them with more appropriate solutions.

annotations/text_upload.py
tjquinn1: It was there when I found it and did not catch it when i was looking it over. Ill move it to its place at the top.

tjquinn1: tried moving this and it creates a circular import

annotations/amphora_utils.py
```diff
@@ -0,0 +1,15 @@
+import requests
+from repository import auth
+"""
+Amphora helper functions.
+"""
+
+def retrieve_amphora_text(url, user):
+    text_request = requests.get(url, headers=auth.jars_github_auth(user))
```
jdamerow, line 8: it's very likely this call would throw an exception at some point. The clean way to handle this would probably  be to catch any thrown exception here and throw a new Amphora exception so the calling code knows there has been an issue.

annotations/tasks.py
```diff
@@ -278,3 +269,78 @@ def handle_file_upload(request, form):
...
+            creation_reader.next()
+            total_count_reader.next()
+            total_count = 0
+            for row in total_count_reader:
+                total_count += 1
+            count = 0
+            task = ImportTasks.objects.create(
+                user = user,
+                file_name = name,
+                total_rows = int(total_count)
```
jdamerow, line 290: is total_count not an int?

annotations/tasks.py
```diff
@@ -278,3 +269,78 @@ def handle_file_upload(request, form):
...
+            )
+            for row in creation_reader:
+                # Store url so we can skip making the request to fetch the json if the urls are the same
+                stored_url = ''
+                try:
+                    parent = Text.objects.get(uri=row["Occurs"])
+                    text = Text.objects.get(part_of_id=parent.id)
+                    occur = text
```
jdamerow, line 298: why don't you assign the text to ```occur``` right away?

annotations/tasks.py
```diff
@@ -278,3 +269,78 @@ def handle_file_upload(request, form):
...
+                except:
+                    # Only make the request and get the json if the url has changed.
+                    url = settings.AMPHORA_RESOURCE_URL + row["Occurs"] + "&format=json"
+                    if stored_url != url:
+                        stored_url = url
+                        text_json, text_content = retrieve_amphora_text(url, user)
+                    add_text = add_text_to_project(user, 1, text_json['id'],project_id)
+                    repo_text_content = repository_text_content(user, 1, text_json['id'], text_content, part_of_id, project_id)
+                    if add_text or repo_text_content is False:
```
jdamerow, line 307: I think your logic is wrong here? what are you trying to test? if at least one of them is False? if so, I believe this is wrong. ```if False or True is False``` would be false; default parentheses would be ```if (add_text) or (repo_text_content is False)```. Also no need for explicit ```False```s. Just do ``` if variable``` or ```if not variable``` for boolean variables.

annotations/tasks.py
jdamerow: you probably want to use ```finally``` here to make sure this is always executed

jdamerow (changes requested):

PR #137, opened 2019-03-18
Story/vgnwb 257

tjquinn1:
submit all appellations in a text with a quadruple [document - contains - appellation]. 
Things to note:
-  a good chunk of this pull request consist of reformatting.
-  The hardcoding of the data object in the relationtemplate_views.py is not good practice but works for the time being and can be refactored later to make the whole process more general.
-  Im not sold on the layout of the whole thing but this was the best I could come up with.
-  Some things are not done in the most efficient way such as using the store to trigger methods instead of using $refs due to the current project configuration.
- for some reason there seems to be a lag when updating the store after the select all button is clicked. I believe the lag only effects the vue dev tools ui from what I could tell and generally users will select templates and concepts which should give it time to catch up if needed and only occurs when the appellations array is large. It should not be a problem but if appellations are missing this could be the culprit.


annotations/static/annotations/js/annotators/appellationlist.js
```diff
@@ -1,101 +1,283 @@
...
+			var minutes = String(date.getMinutes());
+			if (minutes.length == 1) {
+				minutes = '0' + minutes;
+			}
+
+			var day = date.getDate();
+			var monthIndex = date.getMonth();
+			var year = date.getFullYear();
+
+			return day + ' ' + monthNames[monthIndex] + ', ' + year + ' at ' + date.getHours() + ':' + minutes;
```
jdamerow: I know this code is not new and used in other places, but does there seem to be any reason not to use a js library like moment.js? https://momentjs.com/

annotations/views/relationtemplate_views.py
```diff
@@ -201,15 +202,76 @@ def create_from_relationtemplate(request, template_id):
...
+                '3',
+                'project':
+                '1',
+                'start':
+                None,
+                'createdBy':
+                '2',
+                'occur':
+                None
+            }
```
jdamerow, line 267: not sure I understand what's happening here, let's talk about it tomorrow.

jdamerow (dismissed):

tjquinn1:
Removed hardcoded ids and added moment.js

annotations/static/annotations/js/annotators/relationlist.js
```diff
@@ -15,17 +15,32 @@ RelationListItem = {
...
+        },
+        getCreatorName: function (creator) {
+            if (creator.id == USER_ID) {
+                return 'you';
+            } else {
+                return creator.username;
+            }
+        },
+        getFormattedDate: function (isodate) {
+            return moment(isodate).format('dddd') + ' ' + moment(isodate).format('LL') + ' at ' + moment(isodate).format("LT");
```
jdamerow: can't you just do?
```moment(isodate).format('dddd LL [at] LT')```

File: annotations/views/relationtemplate_views.py [hunk 4]
```diff
@@ -201,15 +202,76 @@ def create_from_relationtemplate(request, template_id):
...
 
+def create_from_text(request, template_id):
+    if request.method == 'POST':
+        data = json.loads(request.body)
+        appellations = data['appellations']
+        text_appellation = data['textAppellation']
+        template = get_object_or_404(RelationTemplate, pk=template_id)
+        text = get_object_or_404(Text, pk=data['occursIn'])
+        project_id = data.get('project')
+        if project_id is None:
```
jdamerow, line 224: ```if not project_id```? or could project_id be empty string and that would be valid?

annotations/views/relationtemplate_views.py
```diff
@@ -201,15 +202,76 @@ def create_from_relationtemplate(request, template_id):
...
+                               'appellation': appellation,
+                               'part_field': 'object',
+                               'description': '',
+                               'concept_label': None,
+                               'evidence_reqired': True,
+                               'label': 'Concept',
+                               'part_id': data['part_id'],
+                               'type': 'TP',
+                               'concept_id': None
+                           }],
```
jdamerow: indentation seems off here?

jdamerow:
regarding the layout:
image
- let's make select all, and deselect all icons with titles rather than buttons to save space
- give the dropdown box the bootstrap class
- let's call the "select concept" button "select text concept"
- what does "hide all" and "show all" do? it doesn't seem to change anything.
- if no template is selected and submit is clicked it seems to either not say anything or say "assignment failed". Let's validate before submitting and tell the user what they have to do ("Please select a template", "Please select which concept represents this text.", or "Please select at least one appellation for submission").

The actual functionality seems to work great though! :)

annotations/static/annotations/js/annotators/relationlist.js
tjquinn1: Indeed you can.

annotations/views/relationtemplate_views.py
tjquinn1: This is the work of the good old formatter, I will change to something not so strange.

annotations/views/relationtemplate_views.py
tjquinn1: I did not write this part of the method and figured if it aint broke don't fix so I can not say for sure. From what I can tell I do think that it is possible for it to be a string in some cases. When the project id is initially set in the vue js data properties it is set as a string so if for some reason it did not have a value I am assuming that it would set an empty which then could be submitted elsewhere.

tjquinn1:
The hide all and show all buttons should hide or show all the appellations in the text, this was already of the appellation list functionality. I can remove it from both or add v if's to the template and have them only display when the appellation list is displayed which would declutter it. I should have tried this on a smaller screen to get a better idea of the clutter. I have moved some stuff around and testing using chrome dev tools it looks to be less cluttered on a smaller screen without removing the hide/show all toggles.

jdamerow:
hm, on smaller screens it doesn't seem to fit:
image

tjquinn1:
I added a new css break point called xl to support screens with resolutions over 2000px width. This allows for easier formatting since you do not have to try and shoe horn both imac and medium dpi screens into the same layout. I refactored the layout for both lg and xl layouts so they should both work well. I tested it with the medium dpi setting which is lower than the average macbook pro settings so it should work for everything.

jdamerow:
there is something wrong with the validations; the first time I try to submit something, the validation messages show up fine (if I don't select a concept or template). Afterwards though (after I've submitted some appellations), the messages sometimes don't show up (clicking Submit just doesn't do anything), and once I actually was able to submit the form (although I'm not sure if it created the relations).

And when I select the appellation I've created previously to represent the text, the appellation is highlighted yellow in the list, but clicking Submit doesn't do anything. 

And the appellations previously created for a text, do not show up in the text anymore after I reload the page (the first word of the text is highlight though, so maybe that is part of the problem).

tjquinn1:
Fixed validation errors and removed array reference causing some other errors. Added functionality to allow users to click on an appellation in the text to select/deselect it in the appellation list.

jdamerow:
It looks like after I create a text concept, the button to change that disappears. Is that right? What if I want to change it?

tjquinn1:
Since I used the existing appellation picker functionality it creates an appellation for the text using that concept once you hit submit. I believe there is a ticket to edit appellations which would solve this problem. So yes right now that is correct but not ideal. I should be able to change it after I get these two tickets taken care of.

PR #139, opened 2019-04-12
Story/vgnwb 258

tjquinn1:
Delete appellations that have no relations associated with them. Spinner displays during the delete process and spinner also display when the appellation list/ mass assignment list is loading.


annotations/static/annotations/js/annotators/text.js
```diff
@@ -1547,6 +1552,20 @@ Appellator = new Vue({
                 appellation.selected = (dateappellation_ids.indexOf(appellation.id) > -1);
             });
         },
+        gatherAppellationIdsInRelationset: function (relations) {
+            console.log('runs');
```
jdamerow: remove

annotations/views/rest_views.py
```diff
@@ -292,7 +294,27 @@ def get_queryset(self, *args, **kwargs):
             queryset = queryset.filter(project_id=project_id)
         if position_type and position_type in DocumentPosition.TYPES:
             queryset = queryset.filter(position__position_type=position_type)
-        return queryset.order_by('-created')
+
+        quesryset = queryset.order_by('-created')
```
jdamerow: queryset is misspelled; might create issues down the line (you return queryset)

jdamerow (dismissed):

jdamerow:
where is the delete functionality shown?

annotations/static/annotations/js/annotators/text.js
tjquinn1: removed

annotations/views/rest_views.py
tjquinn1: fixed

tjquinn1:
It is shown on the appellation list. The trashcan denotes that it can be deleted. 
Screen Shot 2019-05-10 at 1 10 54 PM

jdamerow:
There is a bug after creating a new relation. If I create a new appellation during relation creation, I get the delete icon on the appellation and can delete it. If I reload the page, it's fine.

jdamerow:
There is still a bug. When I delete an appellation (not one I've just created, but one I've created previously), the appellation is deleted but a spinning icon is shown for the appellation next in the list (see images before deleting Geneva and after hitting delete):
image
image
Actually, it seems to happen for appellations that are not first or last in the list. And it seems to resort the appellations.

jdamerow:
Also:
- when I create a new appellation and the list of appellations is open, it will have trash icon (correct)
- I then create a new relation that uses that appellation
- after the relation has been created, the trash icon is still there and I can delete the appellation (incorrect)
- if I click to see all relations and then go back to all appellations, it has updated and the trash icon is gone.

PR #140, opened 2019-05-01
Story/vgnwb 248

tjquinn1:
Added download CSV feature. Currently, the functionality allows users to select one or multiple texts to export the corresponding appellations. The created CSVs are saved using file_field so that we can keep track of whos downloading what and only show them previous exports that are their own. I tried about zillions ways to download the CSV and then redirect to the previously downloaded list of CSVs page. I could not make this work in any reliable fashion, it appears that Django cannot handle a download response and a redirect response at the same time. It currently creates the CSV and redirects the user to the previously download CSV list where they have to initiate the download.


annotations/views/download_appellation_views.py
```diff
@@ -0,0 +1,79 @@
...
+@login_required
+def available_csvs(request):
+	'''
+	display list of generated csv that are available for a user to download
+	'''
+
+	template = "annotations/csv_download.html"
+	context = {}
+	csv_list = CsvDownloadList.objects.filter(user=request.user).order_by('-created')
+	print(csv_list)
```
jdamerow, line 66: remove

annotations/views/download_appellation_views.py
```diff
@@ -0,0 +1,79 @@
...
+	context['csvs'] = csv_list
+	return render(request, template, context)
+
+login_required
```
jdamerow, line 70: are you missing and '@' here?

annotations/views/download_appellation_views.py
```diff
@@ -0,0 +1,79 @@
...
+
+@login_required
+def export_appellation_list(request):
+	'''
+	Display list of texts to download appellations from
+	'''
+	template = "annotations/export_appellations.html"
+	context = {}
+	# get texts that are either public or belong to the user and have a content_type of 'text/plain'
+	texts = Text.objects.filter((Q(public=True) | Q(addedBy=request.user)) & Q(content_type='text/plain'))
```
jdamerow, line 21: shouldn't the files be of type text/csv?

jdamerow (changes requested):

PR #141, opened 2019-07-11
Task/vgnwb 259

ghost:
Port to Python 3.


tjquinn1:
When opening a pull request you need t

tjquinn1:
When opening a pull request you need to check if there are merge conflicts and if there are you need to fix them. I pulled this I tried it locally and have had some problems. The concept power package needs to integrated in the project and upgraded to work with python 3. There is also an incompatibility issue with django 1.11 and python 3.7. The issue was fixed in later versions of django, I upgrade django to 2.2 and it fixed that particular issue. I will push that change for you on this branch.

PR #142, opened 2019-08-05
Story/vgnwb 259 3

tjquinn1:
This is a different branch than the previously closed pull request. I forked the conceptpower api repo and converted it to py2/py3 and republished the package. I added the requirements to this project and tested it.


tjquinn1:
Two things about this ticket. First I was getting the following socket error 
Screen Shot 2019-08-06 at 3 12 25 PM . When i looked it up it could be caused by a variety of reason. One is that it is receiving a too many requests which I believe could be from the update of iterm2 that I did. I figured I would let you try it out and see if it is a problem with on your machine. Second I believe the build is failing because travis is not setup for python 3, is that correct? I will leave it open for now so that it gets looked at

PR #143, opened 2019-08-14
Story/vgnwb 259 4

tjquinn1:


jdamerow:
celery version is wrong and doesn't work with python 3.7. when I try to access a users profile page I get an error.

PR #156, opened 2020-01-07
[VGNWB-267] Annotation, Dashboard and User views

nowke:
**Rest endpoints for the following pages:**

- Dashboard
- Users (Contributors): List and detail
- Annotations (Data > Annotations)

Related frontend - https://github.com/diging/vogon-web-client/pull/12


annotations/models.py
```diff
@@ -641,7 +641,7 @@ def dateRepresentation(self):
             _rep += '-' + str(self.day).zfill(2)
         return _rep
 
-    def __unicode__(self):
+    def __str__(self):
```
tjquinn1, line 644: Why did you change these from unicode to str?

annotations/admin.py
```diff
@@ -15,13 +15,13 @@ class VogonUserAdmin(UserAdmin):
...
     # that reference specific fields on auth.User.
-    list_display = ('username', 'full_name', 'email', 'affiliation', 'is_admin')
+    list_display = ('username', 'full_name', 'email', 'affiliation', 'is_admin', 'imagefile')
     list_filter = ('is_admin',)
     fieldsets = (
         (None, {'fields': ('email', 'password')}),
         ('Personal info', {
             'fields': ('full_name', 'affiliation', 'location', 'link',
-                       'conceptpower_uri')
+                       'conceptpower_uri', 'imagefile')
```
tjquinn1, line 24: What is imagefile and what is it used for? Is it an avatar?

tjquinn1 (changes requested):

annotations/models.py
nowke: Its the change from Python 2 to Python 3. 
Changelog - https://docs.djangoproject.com/en/1.11/topics/python3/#str-and-unicode-methods

annotations/admin.py
nowke: Yeah. This is for avatar URL. It has been there in `models.py`, not visible in Django admin.

annotations/models.py
tjquinn1: Got it, it interesting that is didn't automatically change all __unicode__ methods to __str__ methods when we ran 2to3.

tjquinn1 (approved):

PR #157, opened 2020-01-10
[VGNWB-284] Create relation fixes

nowke:


annotations/views/relationtemplate_views.py
```diff
@@ -50,6 +50,22 @@ def list(self, request):
...
+        
+        project_id = data.get('project', None)
+        if project_id is None:
+            project_id = VogonUserDefaultProject.objects.get(
+                for_user=request.user).project.id
+        
+        relationset = relations.create_relationset(
+            template, data, request.user, text, project_id
+        )
+        return JsonResponse({ 'relationset': relationset.id })
```
tjquinn1: Lets change the name to 'relationset_id' so that its clear that its just the id and not an object.

tjquinn1 (changes requested):

annotations/views/relationtemplate_views.py
nowke: Yeah. Made the change

tjquinn1 (approved):

PR #169, opened 2020-02-18
[vgnwb-311] Fixed add text bug where network request was failing

Rudi114:


tjquinn1:
You need to resolve the branch conflicts. When ever you make a pull request you need to resolve the conflicts if there are any.

PR #178, opened 2020-02-24
[VGNWB-320] added delete functionality for texts from projects

Rudi114:


annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
         serializer = ProjectSerializer(project)
         return Response(serializer.data)
 
+    @action(detail=True, methods=['POST'], url_name='removetext')
+    def remove_text(self, request, pk=None):
```
tjquinn1, line 64: Lets call it delete text

annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
         serializer = ProjectSerializer(project)
         return Response(serializer.data)
 
+    @action(detail=True, methods=['POST'], url_name='removetext')
```
tjquinn1, line 63: This not a post request, its a delete request so lets use the right convention

annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
...
+        resource = manager.resource(id=int(text_id))
+
+        defaults = {
+            'title': resource.get('title'),
+            'created': resource.get('created'),
+            'repository': repository,
+            'repository_source_id': text_id,
+            'addedBy': request.user,
+        }
+        text, _ = Text.objects.get_or_create(uri=resource.get('uri'),
```
tjquinn1, line 80: Why are you using get_or_create here when we are trying to delete something? Its alright to copy code but you should understand what its doing and refactor where appropriate not just a rip off of other code.

annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
...
+    def remove_text(self, request, pk=None):
+        text_id = request.data['text_id']
+        repo_id = request.data['repository_id']
+
+        repository = get_object_or_404(Repository, pk=repo_id)
+        project = get_object_or_404(TextCollection, pk=pk)
+        manager = repository.manager(request.user)
+        resource = manager.resource(id=int(text_id))
+
+        defaults = {
```
tjquinn1, line 73: None of defaults are required. What defaults would be needed to delete something? You only need an id. See comment below about copying code.

annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
...
+            'title': resource.get('title'),
+            'created': resource.get('created'),
+            'repository': repository,
+            'repository_source_id': text_id,
+            'addedBy': request.user,
+        }
+        text, _ = Text.objects.get_or_create(uri=resource.get('uri'),
+                                             defaults=defaults)
+        project.texts.remove(text)
```
tjquinn1, line 82: Where is remove coming from? Did you define it somewhere?

annotations/views/project_views.py
```diff
@@ -60,6 +60,30 @@ def add_text(self, request, pk=None):
...
+        
+        serializer = ProjectSerializer(project)
```
tjquinn1, line 84: Why is there a serializer, what are you serializing?

tjquinn1 (changes requested):

annotations/views/project_views.py
Rudi114: .remove is a django function. I found it in the docs https://docs.djangoproject.com/en/3.0/ref/models/relations/

tjquinn1: I did some reading on remove because I was curious. Remove only removes the relationship. So if object bar is related to object foo and you use remove on bar then both foo and bar still exist but bar is no longer related to object foo. While this creates part of the functionality that you intended it does not actually do want you want which is removing it from the database.

PR #179, opened 2020-02-25
[VGNWB-320] Added delete funtionality for texts in

Rudi114:


annotations/views/project_views.py
```diff
@@ -60,6 +60,19 @@ def add_text(self, request, pk=None):
...
+    @action(detail=True, methods=['DELETE'], url_name='deletetext')
+    def delete_text(self, request, pk=None):
+        text_id = request.data['text_id']
+
+        submitted = RelationSet.objects.filter(occursIn_id=text_id, submitted=True)
+        if submitted:
+            return Response(status=status.HTTP_412_PRECONDITION_FAILED)
+        
+        project = get_object_or_404(TextCollection, pk=pk)
+        project.texts.remove(text_id)
```
tjquinn1: From your other PR:

I did some reading on remove because I was curious. Remove only removes the relationship. So if object bar is related to object foo and you use remove on bar then both foo and bar still exist but bar is no longer related to object foo. While this creates part of the functionality that you intended it does not actually do want you want which is removing it from the database.

annotations/views/repository_views.py
```diff
@@ -73,7 +73,7 @@ def retrieve(self, request, pk=None, repository_pk=None):
         try:
             master_text = Text.objects.get(uri=result.get('uri'))
         except Text.DoesNotExist:
-            master_text = None
+            master_text = Text.objects.create(uri=result.get('uri'),title=result.get('name'),public=result.get('public'),content_type=result.get('content_types'),repository_source_id=result.get('id'),repository_id=repository_pk,addedBy_id=1)
```
tjquinn1, line 76: What is this doing? Is this from a different ticket? My intuition says you probably shouldn't be creating something when your trying to delete something.

tjquinn1 (changes requested):

annotations/views/project_views.py
Rudi114: Thanks for the info. Ill change that here and keep it in mind for the future.

annotations/views/repository_views.py
Rudi114: This was from ticket 311. It had to do with adding text. I made the same edit here as it was making it difficult to test the delete functionality without the add text button working.

Rudi114:
Now instead of showing red button, it should show greyed out button with tool tip sayng that "Texts cannot be removed after annotations have been submitted". I thought this was better as it is less likely to leave the user confused as to why they cant delete. Let me know if you want me to take away the button all together.

annotations/views/repository_views.py
```diff
@@ -73,12 +73,21 @@ def retrieve(self, request, pk=None, repository_pk=None):
...
         aggregate_content = result.get('aggregate_content')
+
+        submitted = False
+        i = 0
+        while i < len(master_text.children):
```
tjquinn1: This is not a great way to do this. One it is very inefficient and seems to have created long load times. Two its not as readable as it could be and has added unneeded lines of code. Use a for loop to iterate over children and then add a break statement after submitted is changed to true. There is no point continuing to iterate over the annotations once the condition is meet, it just wastes time and resources.

tjquinn1 (changes requested):

tjquinn1:
resolve conflicts in front end

annotations/views/project_views.py
```diff
@@ -60,6 +60,19 @@ def add_text(self, request, pk=None):
         serializer = ProjectSerializer(project)
         return Response(serializer.data)
 
+    @action(detail=True, methods=['DELETE'], url_name='deletetext')
```
tjquinn1: There is no need to create a custom action here. Viewsets give us a destroy action by default. You can see a list of default actions here https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions. You can override the default destroy method to implement what you need. This way we do not have unneeded routes and we are leveraging the full usefulness of Viewsets.

annotations/views/project_views.py
```diff
@@ -60,6 +60,19 @@ def add_text(self, request, pk=None):
...
+        text_id = request.data['text_id']
+
+        submitted = Appellation.objects.filter(occursIn_id=text_id, submitted=True)
+        if submitted:
+            return Response(status=status.HTTP_412_PRECONDITION_FAILED)
+        
+        project = get_object_or_404(TextCollection, pk=pk)
+        project.texts.filter(pk=text_id).delete()
+
+        return Response(status=status.HTTP_200_OK)
```
tjquinn1: Also lets return a 204 since it is successful and we are not returning anything. See this stack post for more clarification if you are interested. https://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete

annotations/views/project_views.py
```diff
@@ -60,6 +60,19 @@ def add_text(self, request, pk=None):
...
 
+    @action(detail=True, methods=['DELETE'], url_name='deletetext')
+    def delete_text(self, request, pk=None):
+        text_id = request.data['text_id']
+
+        submitted = Appellation.objects.filter(occursIn_id=text_id, submitted=True)
+        if submitted:
+            return Response(status=status.HTTP_412_PRECONDITION_FAILED)
+        
+        project = get_object_or_404(TextCollection, pk=pk)
```
tjquinn1, line 70: It wouldn't let me add a comment in the front end but since we could possible return a 404 lets handle that in the front end as well. Right now we are only handling the 412.

tjquinn1 (changes requested):

tjquinn1 (approved):

PR #180, opened 2020-03-02
[VGNWB-314] Add password reset views

nowke:
Views for

- Forgot password - Sends email in the console
- Reset password


tjquinn1:
Resolve conflicts and is there a client a ticket for this as well?

nowke:
Corresponding ticket - https://diging.atlassian.net/browse/VGNWB-314

tjquinn1:
Resolve conflicts

accounts/views.py
```diff
@@ -73,6 +76,81 @@ def citesphere_token(request):
...
+        username = request.data.get('username', None)
+        if username:
+            try:
+                user = VogonUser.objects.get(username=username)
+            except VogonUser.DoesNotExist:
+                return Response({ "success": False, "message": "User not found" }, status=404)
+            token = jwt.encode({ 
+                'exp': arrow.utcnow().shift(days=1).timestamp,
+                'username': user.username,
+            }, settings.SECRET_KEY, algorithm='HS256').decode('UTF-8')
```
tjquinn1, line 90: why do you encode the secret key here?

accounts/views.py
```diff
@@ -73,6 +76,81 @@ def citesphere_token(request):
...
+            reset_token = ResetToken(user=user, token=token)
+            reset_token.save()
+
+            # Email token
+            send_mail(
+                'Vogon password reset',
+                f'Reset link: http://localhost:8080/reset-password/{token}',
```
tjquinn1: lets make the reset link and the vogon email be configurable from the settings file

accounts/views.py
```diff
@@ -73,6 +76,81 @@ def citesphere_token(request):
...
+                'test@vogonweb.com',
+                [user.email],
+                fail_silently=True,
```
tjquinn1: do we want this to fail silently?

accounts/views.py
```diff
@@ -73,6 +76,81 @@ def citesphere_token(request):
...
+            )
+
+            return Response({ "success": True })
+        return Response({ "success": False, "message": "Username not specified" }, status=400)
+
+class ResetPasswordView(APIView):
+    def post(self, request):
+        username = request.data.get('username', None)
```
tjquinn1: I think we should use a serializer for this. That way we can validate the data

vogon/settings.py
```diff
@@ -62,6 +62,7 @@
     "django.middleware.security.SecurityMiddleware",
 )
 
+EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
```
tjquinn1: lets make this an env variables

tjquinn1 (changes requested):

accounts/views.py
nowke: JWT uses secret key to encode the contents of the dict.

accounts/views.py
nowke: Fixed

accounts/views.py
nowke: Done

vogon/settings.py
nowke: Done

tjquinn1:
Make it so, Jenkins.

PR #181, opened 2020-03-06
[VGNWB-254] Add appellation edit view

nowke:
- Add support for editing an appellation

Related FE - https://github.com/diging/vogon-web-client/pull/30


tjquinn1:
See notes on frontend.

PR #191, opened 2020-08-10
[VGNWB-326] Add graph visualiztion for texts

nowke:
Corresponding Frontend PR - https://github.com/diging/vogon-web-client/pull/52


nowke:
Reopening as discussed.

PR #210, opened 2021-07-29
Story/vgnwb 350

sudheerad9:

Fix dates appellations 
  
... describe here ...


annotations/views/amphora_views.py
```diff
@@ -115,8 +115,8 @@ def retrieve(self, request, pk=None, repository_pk=None):
         aggregate_content = result.get('aggregate_content')
 
         submitted = False
-        for child in range(len(master_text.children)):
-            if Appellation.objects.filter(occursIn_id=master_text.children[child], submitted=True):
+        for index, child in enumerate(master_text.children):
```
tjquinn1: if you are not using the index why do you use enumerate here? Why not just loop over the text?

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+
+from annotations.models import DateAppellation, DocumentPosition
+from annotations.serializers import DateAppellationSerializer, DocumentPositionSerializer
+
+class DateAppellationViewSet(viewsets.ModelViewSet):
+    queryset = DateAppellation.objects.all()
+    serializer_class = DateAppellationSerializer
+
+    def create(self, request):
+        data = request.data.copy()
```
tjquinn1: why do you copy data?

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+        position = data.pop('position', None)
+        if 'month' in data and data['month'] is None:
+            data.pop('month')
```
tjquinn1: if you don't need to save the value into another variable I wouldn't use pop, just access it by key. That way your not mutating the original data unnecessarily.

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+        
+        serializer = DateAppellationSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        instance = serializer.save()
+        
+        text_id = serializer.data.get('occursIn')
+        if position:
+            if not isinstance(position, DocumentPosition):
+                position_serializer = DocumentPositionSerializer(data=position)
+                position_serializer.is_valid(raise_exception=True)
```
tjquinn1: Should this exception be handled?

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+                position = position_serializer.save()
+
+            instance.position = position
```
tjquinn1: Why not just use `position_serializer.save()` instead of saving it into a variable just to save into another variable.

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            reserializer.data, 
+            status=status.HTTP_201_CREATED,
+            headers=headers
+        )
+    
+    def partial_update(self, request, *args, **kwargs):
+        instance_object = self.get_object()
+        data = request.data.copy()
```
tjquinn1: same as above

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+        if 'day' in data and data['day'] is None:
+            data['day'] = 0
+            
+        serializer = DateAppellationSerializer(instance_object, data=data, partial=True)
+        serializer.is_valid(raise_exception=True)
+        instance = serializer.save()
+        if position:
+            if not isinstance(position, DocumentPosition):
+                position_serializer = DocumentPositionSerializer(instance_object.position, data=position, partial=True)
+                position_serializer.is_valid(raise_exception=True)
```
tjquinn1, line 66: Same as above, does this need to be handled?

annotations/views/date_appellation_views.py
```diff
@@ -0,0 +1,74 @@
...
+                position = position_serializer.save()
+
+            instance.position = position
```
tjquinn1: same as above

tjquinn1 (changes requested):

tjquinn1:
Make it so, Jenkins.

tjquinn1:
Make it so, Jenkins.

PR #217, opened 2021-11-12
Bug/vgnwb 368

sudheerad9:

fix signup error
  
... describe here ...

### DIFF
--- a/annotations/views/user_views.py
+++ b/annotations/views/user_views.py
@@ -18,235 +18,234 @@ from rest_framework.decorators import action, api_view
 from accounts.views import VogonUser
 
 from annotations.models import (VogonUser, Text, Appellation, RelationSet,
-                                Relation)
+								Relation)
 from annotations.serializers import (
-    ProjectSerializer, TextSerializer, RelationSetSerializer,
-    UserSerializer, TextCollectionSerializer
+	ProjectSerializer, TextSerializer, RelationSetSerializer,
+	UserSerializer, TextCollectionSerializer
 )
 from accounts.serializers import UserSerializer as VogonUserSerializer
 
 
 
 class VogonUserAuthenticationForm(AuthenticationForm):
-    class Meta:
-        model = VogonUser
+	class Meta:
+		model = VogonUser
 
 
 class UserViewSet(viewsets.ModelViewSet):
-    """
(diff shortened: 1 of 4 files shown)


annotations/views/user_views.py
```diff
@@ -227,3 +229,22 @@ def get_weekly_annotations(self, user):
 				'count': weekly_annotations.get(week.datetime, 0)
 			})
 		return result
+	
+
```
tjquinn1: remove extra line

annotations/views/user_views.py
```diff
@@ -227,3 +229,22 @@ def get_weekly_annotations(self, user):
...
+@api_view(["POST"])
+@authentication_classes([])
```
tjquinn1: is this necessary if you are using no classes?

annotations/views/user_views.py
```diff
@@ -227,3 +229,22 @@ def get_weekly_annotations(self, user):
...
+@permission_classes([])
```
tjquinn1: same as above

annotations/views/user_views.py
```diff
@@ -227,3 +229,22 @@ def get_weekly_annotations(self, user):
...
+def create(request):
+	try:
+		user = VogonUser.objects.create(**request.data)
+		user.save()
+		refresh = RefreshToken.for_user(user)
+		return Response({
```
tjquinn1, line 239: This should also return a response status 201 if the created succeeded

tjquinn1 (changes requested):

annotations/views/user_views.py
```diff
@@ -227,3 +229,19 @@ def get_weekly_annotations(self, user):
 				'count': weekly_annotations.get(week.datetime, 0)
 			})
 		return result
+	
+@api_view(["POST"])
+def create(request):
```
tjquinn1, line 234: Is this necessary? Whats wrong with /users?

sudheerad9: Seperating user creation and signup. In user creation we need not return the access and refresh tokens, for signup UI is setting those in local storage. So the signup API creates user and then returns the refresh token for users.

jophals (approved):
Approved for merge.

PR #221, opened 2022-01-27
Story/vgnwb 371

sudheerad9:

...Put description here...
  
... describe here ...

### DIFF
--- a/annotations/views/project_views.py
+++ b/annotations/views/project_views.py
@@ -41,7 +41,9 @@ class ProjectViewSet(viewsets.ModelViewSet):
     def create(self, request):
         request.data['createdBy'] = request.user
         request.data['ownedBy'] = request.user
+        participants = request.data.pop('participants')
         project = TextCollection(**request.data)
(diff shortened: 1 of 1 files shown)


annotations/views/project_views.py
```diff
@@ -129,6 +130,15 @@ def list_user_projects(self, request):
...
         return Response(serializer.data)
     
+    @action(detail=False, methods=['get'], permission_classes=[IsAdminUser])
+    def list_all_projects(self, request):
+        response_data = {}
+        projects = TextCollection.objects.all()
+        serializer = ProjectSerializer(projects, many=True)
+        response_data["results"] = serializer.data
+        response_data["count"] = len(serializer.data)
+        return Response(response_data)
```
tjquinn1: return a status response.

tjquinn1 (changes requested):

tjquinn1:
reopen with FE

jophals (approved):
Approved

PR #230, opened 2023-11-07
Story/vgnwb 366

jophals:

retrieve API to fetch item from citesphere, make the text display flow work for citesphere like for amphora
  
main ticket items already addressed in different pull request, these changes are minor fixes and refactorings


annotations/model_utils.py
```diff
@@ -0,0 +1,8 @@
+import re
+
+def help_text(text):
```
jdamerow: have you tried if you really need this method?

annotations/relations.py
```diff
@@ -521,3 +522,86 @@ def update_relationset(template, raw_data, creator, text, project_id=None, relat
...
+                'part_field': pred
+            }
+            relation_data[dkey] = create_appellation({}, payload, project_id=project_id, creator=creator, text=text)
+
+def _get_cache(template_part, cache={}):
+    if cache != None:
+        key = template_part.id 
+        if key in cache:
+            return cache[key]
+    return None
```
jdamerow: this could be a one liner:
`return cache[template_part.id] if cache and template_part.id in cache else None`

And two other things:
- `_get_cache` implies this method returns a cache. but it returns a cached object (big difference)
- since `template_part` is never used except to get the id, why don't you just pass the id (which would make it easier to read).

annotations/relations.py
```diff
@@ -521,3 +522,86 @@ def update_relationset(template, raw_data, creator, text, project_id=None, relat
...
+    relation_data = _get_relation_data(relationset, creator, text)
+    _set_relation_data(template_part, provided, relationset, relation_data, field_handlers, inspect.currentframe().f_code.co_name, cache, appellation_cache, project_id, creator, text)
+    relation = Relation.objects.create(**relation_data)
+
+    if cache != None:
+        cache[template_part.id] = relation
+    return relation
+
+def _update_relation(template_part, provided, relationset, cache={}, appellation_cache={}, project_id=None, creator=None, text=None, required=None, _as_key=None):
+    relation_cache = _get_cache(template_part, cache)
```
jdamerow: the method actually returns the cached relation not the relation cache, doesn't it?

annotations/relations.py
```diff
@@ -521,3 +522,86 @@ def update_relationset(template, raw_data, creator, text, project_id=None, relat
...
+    if relation_cache != None:
+        return relation_cache
+
+    field_handlers = _get_field_handlers(appellation_cache, project_id, creator, text, required, _as_key)
+    relation_data = _get_relation_data(relationset, creator, text)
+    _set_relation_data(template_part, provided, relationset, relation_data, field_handlers, inspect.currentframe().f_code.co_name, cache, appellation_cache, project_id, creator, text)
+
+    relation_id_q = relationset.constituents.all().values('id')
```
jdamerow: It looks like all you need is the first relation, is that right? If so, is there a need to get all ids? Can't you just do `relationset.constituents.all().first()`?

annotations/relations.py
```diff
@@ -521,3 +522,86 @@ def update_relationset(template, raw_data, creator, text, project_id=None, relat
...
+    field_handlers = _get_field_handlers(appellation_cache, project_id, creator, text, required, _as_key)
+    relation_data = _get_relation_data(relationset, creator, text)
+    _set_relation_data(template_part, provided, relationset, relation_data, field_handlers, inspect.currentframe().f_code.co_name, cache, appellation_cache, project_id, creator, text)
+    relation = Relation.objects.create(**relation_data)
+
+    if cache != None:
+        cache[template_part.id] = relation
+    return relation
+
+def _update_relation(template_part, provided, relationset, cache={}, appellation_cache={}, project_id=None, creator=None, text=None, required=None, _as_key=None):
```
jdamerow, line 543: there needs to be documentation what this method is doing

jdamerow (changes requested):

annotations/model_utils.py
jophals: i commented out the method and everything seems fine but I'm not sure that I'm adequately testing it.

jophals (commented):
Implemented fixes

